PR #229, opened 2026-06-04
feat(pcb): make boards movable and track their transform when editing

ecto:
## What

Task 2 of the modeless-PCB follow-up: **a board moved/rotated as a part now edits correctly**. This turned out to need two fixes.

### 1. Boards were unmovable (prerequisite)

A board has scale/rotate/translate wrapper nodes like any part, but it's created via the wasm `create_feature` (raw kind + params, used because the `board` payload is a big JSON string) — which, unlike `add_feature`, **never registered a stable id**. So `set_translation` / `set_rotation` / `set_scale` resolved to nothing and silently no-oped: you literally could not move a board.

- Added `DocumentApi::create_feature_raw` (mirrors `add_feature` — registers the stable id) and routed the wasm `create_feature` through it.
- Any raw-created feature is now transformable, deletable, and updatable by id.

### 2. Editing assumed the origin

The kernel slab bakes the board's transform, but `PcbScene` (copper / footprints / interaction plane) and the auto-fit camera were drawn at local origin — so a moved board's copper detached from its slab.

- Added `getPcbBoardTransform` (reads the board's translate/rotate/scale nodes, mirroring the evaluator's `translate ∘ rotate ∘ scale` compose order and degree→radian convention) and `findPcbBoardPart` in core.
- `ViewportContent` wraps `<PcbScene>` in that transform group (inside the existing Z-up→Y-up rotation group), and `usePcbAutoFit` transforms its framing centroid/vertices/normal so the camera lands on the board's real location.
- Picking already projects via the group's `worldToLocal`, so routing/dragging track the moved board automatically.

## Verification

- ✅ **Boards now move/rotate live**: `setTranslation`→(40,20,-3) and `setRotation`→(0,0,30°) update the board's transform nodes (both were no-ops before).
- ✅ **Kernel slab bakes it correctly**: that board's evaluated slab bbox is `[25,20,-3.8]→[83.3,71,-2.2]` — a 50×30 board rotated 30° about Z (bbox 58.3×51 ✓) and translated (40,20,−3) with the centered z. `PcbScene` consumes the identical transform (same nodes), so copper/picking/camera track by construction.
- ✅ Earlier verified the interaction plane renders coplanar with the board (the #227 `worldToLocal` picking fix), so wrapping it in the transform group keeps picking correct.
- ✅ `cargo test --workspace`, `npm run build -w @vcad/app` (+ core), clippy & fmt on changed crates all pass. New test: `create_feature_raw_is_movable`.

Note: as in #228, reliable framed 3D screenshots weren't achievable in the headless preview (on-demand frameloop + the in-canvas instrumentation wouldn't consistently mount), so tracking was verified by reading the live evaluated transforms rather than pixel capture.

## Follow-ups (not in this PR)
Route-vs-enclosure interference (`useElectronicsSync`) still computes component AABBs in board-local space; for a moved board it should transform them into world space too. Worth a small follow-up.


ecto:
Added the interference follow-up to this branch (2cd7258b): `useElectronicsSync` now maps board-local component bodies into world via the board transform before the overlap test, so a moved/rotated board clashes against the right location. New test `maps board-local bodies into world via boardToWorld`. The transform-awareness is now complete (copper, picking, camera, **and** interference).

PR #253, opened 2026-06-11
feat(sheet): SendCutSend-ready DXF export and parametric bend relief

ecto:
## What

Makes the sheet-metal flat-pattern DXF natively accepted by SendCutSend-class fab services, and adds parametric bend-relief notch generation with a matching DFM rule.

A real upload of the [origami-crane test part](https://github.com/ecto/vcad/blob/claude/nostalgic-wilbur-ebaf66/crates/vcad-kernel-sheet/src/dxf.rs) to SendCutSend failed three ways until hand-repaired: **"open entities"** (17 disjoint per-panel polylines with uncovered allowance strips), **undetected bends** (solid bend lines — their parser keys off the entity linetype), and **"Relief Insufficient"** (no notches where bend ends meet adjacent material). This PR fixes all three in the kernel so the exported file works first try.

### DXF format (`vcad-kernel-sheet`)
- **New `silhouette` module** — merges panel outlines + bend-allowance strips into a single closed CUT profile via directed-edge cancellation (no polygon-boolean dependency; exploits the model invariant that strips share full edges with their panels). Collinear edge-splitting keeps partially-overlapping seams (notched hinges) cancelling correctly.
- **Dashed bend centerlines** — `DASHED` declared in the LTYPE table and applied per entity; bend lines moved from the parent hinge edge to the allowance-strip **midline** (the true bend centerline — previously half an allowance off).
- `FlatPattern` carries `allowance_strips_2d`; the WASM DTO exposes the merged `silhouette_2d`.

### Bend relief (new `relief` module)
- `add_bend_relief` / `add_all_bend_reliefs` cut **V-shaped relief notches into the parent panel outline**, so the flat pattern, DXF, and folded 3D body stay consistent from one source of truth. Defaults: width = thickness, depth = R + t past the bend centerline.
- V-cuts replace the corner vertex. Where two hinges share a body corner (wing/tail roots on a polygonal blank) the two requested notches **merge into one V** — the naive rectangular slots overlap there, which is invalid geometry in 2D and 3D alike (caught by the golden fixture).
- **New DFM rule `sheet.bend_relief`** (Warning) flags un-relieved bend ends using the *same predicate* the fix applies, so fix-then-check always converges to clean. `ShopProfile` gains field-tolerant `relief_depth_mm` (0 = auto R + t). `sheet_metal_suggest_fix` emits a structured `add_bend_relief` action.

### MCP server
- `slimPreviewForInlineUi` no longer strips tool-result payloads for clients that never declared the MCP Apps UI extension. Viewer-less hosts (Claude Code, plain agents) were losing entire results (`get_document`, `sheet_metal_create`) to the `{document_id}` stub — now gated on the client capability from the initialize handshake.

## Testing

- `cargo test --workspace` green; `cargo clippy --workspace -- -D warnings` clean on touched crates (pre-existing cocoa deprecations in the Tauri shell unchanged); `cargo fmt --check` clean.
- Golden fixture: crane topology (hexagonal body, chained head fold, diagonal hinges) asserts exactly one closed CUT silhouette with area conservation, dashed centered bend lines, 8 flagged relief ends, zero after the fix, and silhouette/DXF consistency post-notching.
- MCP package: 39/39 against the rebuilt WASM. End-to-end node run reproduces this session's crane: 1 silhouette loop, dashed `BEND_UP`/`BEND_DOWN` entities structurally matching the hand-repaired file SendCutSend accepted, and `sheet_metal_check` surfacing the 8 relief warnings with structured fixes.
- Rebuilt `packages/kernel-wasm` artifacts are included (checked in by convention).

## Notes for review

- Crease lines in `FlatPattern` changed meaning (hinge edge → bend centerline). All consumers pass through coordinates; the half-allowance shift is the bug fix.
- Relief is kernel-side only for now: exposing it at `sheet_metal_create` time needs a `SheetMetalBendRelief` IR op (vcad-ir → packages/ir → engine → WASM chain) — deliberate follow-up. The suggest-fix text is explicit about this.


ecto:
Consolidated into #252 (per discussion) — that branch is the superset implementation (kernel-native 2D booleans, SCS shop catalog, IR-level `bend_relief`, folded STEP path, property tests). Three ideas from this PR were folded in there as of f3f6d345:

- **`slimPreviewForInlineUi` capability gating** — ported nearly verbatim; viewer-less hosts keep full tool results.
- **`silhouette_2d` on the flat-pattern DTO** (same field name/shape: exterior ring first, then holes) — exposed through the engine's `SheetMetalFlatPattern` type.
- **The shared-corner relief scenario** this PR's V-cuts were designed around — covered in #252 by the boolean-difference notches (overlapping cuts merge into one valid region) and now locked in with two tests: a hinge ending at a 45° chamfer, and two hinges sharing a chamfered corner.

The two relief shapes differ deliberately: #252 keeps rectangular notches (the shape SCS publishes width/depth specs for, and what the task spec called for) cut via 2D boolean difference, which makes the corner-merge property fall out of the geometry rather than special-casing it.

Suggest closing this PR in favor of #252 once it's green.

PR #255, opened 2026-06-12
mecheval: claude-mcp matrix rerun — 228 graded attempts with the verified agent loop

ecto:
## What

The first honest agentic benchmark run since the verified-agent-loop landed in #254 — every prior `claude-mcp` run was effectively blind (image blocks never reached the model). This PR adds the batch tooling that made the run survivable, two bug fixes the run itself surfaced, and **228 graded run blobs** for `claude-mcp-claude-opus-4-7` across tiers a1–a6.

### Tooling
- **`mecheval-matrix`** batch runner: worker pool over (task × attempt) jobs with per-attempt process isolation, wallclock kill timers, retry-with-backoff on transient API errors, a circuit breaker (6 consecutive errors → 5-min pause; 3 trips → clean abort), and `--since` top-up mode that counts existing blobs toward each task's attempt target — interrupted batches resume exactly where they died. All three interruptions this run (API brownout, credit exhaustion, host kill) resumed losslessly.
- Fresh-attempt semantics: pass^k scores the most recent k blobs, so a rerun after a harness fix never lets stale (blind) runs leak into the new number.

### Fixes surfaced by the run
- **Per-block MCP result extraction** (harness): the MCP Apps preview handle appends a second text block to geometry tool results; joining blocks before `JSON.parse` corrupted every extracted `.vcad` with trailing characters. Each text block is now parsed independently (regression-tested).
- **`get_document` stays JSON-pure** (server): tools whose text body is machine-parseable JSON no longer get handle text appended — `structuredContent` only.

## Results so far (partial matrix, 228/321 attempts)

| tier | pass rate |
|------|-----------|
| a1 | 83% (30✓/6✗) |
| a2 | 74% (48✓/17✗) |
| a3 | 56% (25✓/20✗) |
| a4 | 42% (17✓/23✗) |
| a5 | 55% (22✓/18✗) |
| **overall** | **63%** |

Context: the blind agentic loop scored **40%**; single-shot (`claude-direct`, same model) scores **81%**. The sighted loop closed most of the gap, and the failure signature changed qualitatively — almost every failure is `bbox`+`mass_props` together (placement/sizing misreads), not invalid geometry. The kernel essentially never produced a broken solid across 228 attempts.

## Reviewer notes

- **93 attempts remain** (suite F fits, physics, visual, a6 stragglers) — paused for API budget. Resume: `mecheval-matrix --solver claude-mcp-claude-opus-4-7 --since 20260611T170000Z --concurrency 3`. Leaderboard publication should wait for the full matrix.
- Two known harness gaps documented for follow-up, deliberately NOT changed mid-batch to keep runs homogeneous: (1) agents occasionally call `close_document` mid-run, vaporizing the attempt (~3%) — it should be excluded from benchmark tool lists like the oracle tools; (2) task `max_tokens` limits are declared but unenforced (true of all historical runs).
- No prompt caching in the harness — each iteration resends full history + ~45 tool schemas (~178k input tokens for even simple tasks). Worth adding `cache_control` before the next full matrix; it would cut input cost ~90%.


ecto:
## Matrix complete — final numbers

**295 graded attempts across the 52 runnable tasks** (commit 22fa8a7f). The remaining 12 tasks (suites D + F) are structurally unrunnable: their agent-visible reference images were never generated — only the hidden grader geometry exists. Asset-generation pipeline tracked as follow-up work.

### Sighted agentic (`claude-mcp-claude-opus-4-7`, fresh window)

| tier | per-attempt | pass^5 |
|------|-------------|--------|
| a1 | 84% | 4/7 |
| a2 | 74% | 4/13 |
| a3 | 56% | 2/9 |
| a4 | 42% | 1/8 |
| a5 | 55% | 1/8 |
| a6 | 47% | 1/6 |
| c (physics) | 0% | 0/1 |
| **total** | **60% (157/262)** | **13/52 (25%)** |

### The three-way comparison

- **Blind agentic → sighted agentic** on the identical 5 originally-attempted tasks: **40% → 64%**. Eyes + image forwarding + vocabulary are worth ~24 points.
- **Sighted agentic vs single-shot** on shared coverage (a1–a4, the only tiers `claude-direct` ran): **65% vs 81%** per-attempt; pass^5 gap is wider (11/37 vs 32/37). **Single-shot still wins where both are measured.**
- **Coverage**: agentic uniquely covers a5/a6/c (single-shot was never run there); it holds 47–55% on tiers with reference geometry and multi-feature parts.

### Read on the result

The verified loop closed most of the blind-agentic gap but has **not yet beaten single-shot** — the "prove" gate for a public mecheval launch is not met. The dominant failure mode is unchanged (bbox+mass_props together: right shape, wrong size/placement), which suggests the isometric line-art render confirms *shape* but doesn't reveal *dimensional* errors. Plausible next levers before a re-run: orthographic/dimensioned `render_view` views, a system-prompt nudge to verify with `inspect_cad` against the spec before stopping, and investigating why iteration introduces placement errors that one-shot generation avoids (claude-direct hit 100% on a2).

Cost note: the final 93-attempt leg ran on the cached harness — ~16 uncached input tokens per request.

PR #284, opened 2026-06-16
feat: live Receipt ledger in the electronics workspace (#280)

### LINKED ISSUE #280: Receipt: live ledger in the app electronics panel
Follow-up to #276 (the Receipt).

Surface the Receipt in the app's electronics workspace as a live ledger: as the agent (or the user) mutates the board, render each entry — the per-rule before/after delta, the credit/blame attribution, and the verdict badge — using `renderReceiptHtml` (already in `packages/mcp/src/receipt/render.ts`) as the reference design.

This is the "shared canvas" surface from the broader electronics-UX direction: the human watches the agent's work accrue as an auditable, re-runnable ledger rather than a silent stream of mutations. Bigger lift than the headless v0 — needs a React component in the electronics panel + store wiring, and live-verify has known friction (HMR resets the electronics session; rebuilding @vcad/core needs a dev-server restart).

ecto:
Closes #280.

## What

Surface **the Receipt** (shipped in #276 for the MCP server) as a **live ledger** in the app's electronics workspace. As you edit a board, each mutation that changes DRC logs one attributed entry — the per-rule before/after delta, the credit/blame attribution (footprint vs routing), shorts called out, and a verdict badge. A "Ledger" toggle in the electronics header opens a floating panel.

The in-app ledger uses the **same browser-side kernel DRC** the workspace already runs, so it and the server-side `route_nets` receipt attribute identically.

## How

**`refactor(core)`** — move the Receipt engine into `@vcad/core` so both the app and the MCP server share one implementation:
- `packages/core/src/utils/receipt/{types,engine,session,render}.ts` (moved from `@vcad/mcp`).
- `hash.ts` — a small dependency-free synchronous content hash replacing `node:crypto` sha256 (`crypto.subtle` is async and would force `buildEntry` async). The fingerprint is a "same board" check, not cryptographic.
- `adapter.ts` — `snapshotFromViolations()` bridges the flat DRC list to the `DrcSnapshot` the engine consumes.
- `@vcad/mcp` now imports the engine from `@vcad/core`; the `route_nets` integration is unchanged.

**`feat(app)`** — the ledger:
- `electronics-store`: `receiptEntries` + `showReceiptPanel` + a `pendingMutation` label channel.
- `useReceiptRecorder`: debounced — after each settled mutation it runs DRC, diffs via `buildEntry`, appends a non-no-op entry. Board/document switches re-seed + clear. Autoroute tags its burst so the coalesced entry reads "autoroute".
- `ReceiptPanel` (floating, `LengthTunePanel`-styled) + a "Ledger" toggle with count in the header.

## Verification

- `@vcad/core` builds (real `tsc`) + **119 core tests** pass (engine, session, adapter); `@vcad/mcp` **159 tests** pass; app `tsc --noEmit` **clean**.
- Live (dev server): the Ledger toggle and panel render in the electronics workspace, and the recorder runs without errors — correctly emitting **nothing** when an edit doesn't change DRC (placing parts changed ERC, not DRC; ledger stayed empty, confirming the no-op gating). The entry **attribution** itself is locked by the core engine tests + the `route-receipt` integration test, which drives the **real kernel** router and asserts the verdict/shorts.

## Scope (v1)

User-driven mutations only. Agent (MCP) mutations aren't synced live to the app yet (the realtime channel carries CRDT ops between users, not server-side receipts) — the engine now lives in `@vcad/core` so a v2 is purely a transport addition: have the MCP server write each `ReceiptEntry` to a realtime channel and the app append it via the same `appendReceiptEntry` + `ReceiptPanel`.


ecto:
### The Receipt ledger, populated

Receipt ledger — the 4→72 re-route regression

Two `autoroute` entries on a demo board, attributed by the shipped `@vcad/core` engine:

- **#1 — improved-with-regressions** (DRC 8→4): fixed 5 unrouted nets (credit), introduced 1 via hole-to-hole, and the 3 connector pad-pitch faults are tagged **pre-existing · not the agent**.
- **#2 — regression** (DRC 4→72): the re-route silently stacked copper into **10 hard shorts** + 58 trace-clearances — called out in red — while the same 3 footprint faults stay pre-existing. The naive 4→72 headline hides nothing here; the per-rule delta + attribution tells the real story.

This image is the standalone `renderReceiptHtml` (the reference renderer that ships in this PR) so it's legible at a glance. The **in-app docked `ReceiptPanel`** renders the same entries — verified live in the dev server, where its DOM read back exactly:

```
Receipt ledger 2 · Clear ✕
autoroute · Regression · DRC 4 → 72 (+68) · ⚠ 10 hard shorts introduced — board electrically broken
  Clearance 3→61 +58 · Short 0→10 +10 · ✗ 68× routing · 3× footprint · 1× via
autoroute · Improved · regressions · DRC 8 → 4 (-4)
  UnconnectedNet 5→0 -5 · HoleToHole 0→1 +1 · ✓ 5× unrouted · ✗ 1× via · 3× footprint
```

(The app panel is dark-themed and docks bottom-left under a "Ledger" toggle in the electronics header; a raw screenshot of it is too dark to be useful, hence the standalone render above.)

Issue #298, opened 2026-06-17
Kernel: chained curved-surface boolean resurfaces subtracted sphere


ecto:
## Summary

The BRep boolean pipeline (`crates/vcad-kernel-booleans`) produces invalid topology when a **curved tool is subtracted from a solid that already carries a concave curved face from a prior boolean**. A subtracted sphere "resurfaces" as positive geometry.

This was discovered while fixing the loon `let`-drops-body bug (#PR). The loon evaluator and IR conversion compose nested/chained `difference` **correctly** — this is purely a kernel boolean robustness issue.

## Minimal repro (2 ops)

```rust
let body   = Solid::cube(90.0, 60.0, 28.0).fillet(14.0);
let sphere = Solid::sphere(36.0, 0).translate(45.0, 30.0, 58.0);
let cyl    = Solid::cylinder(5.0, 4.0, 0).translate(45.0, 42.0, 25.0);

let result = body.difference(&sphere)  // correct alone: dished cube, z-max ≈ 28
                 .difference(&cyl);     // BUG: sphere resurfaces, z-max ≈ 94
```

Expected: a filleted body with a spherical dish + a small cylindrical notch (bbox z ≈ 0..28).
Observed: the second difference yields invalid topology in which the subtracted sphere reappears as positive geometry (bbox z ≈ 0..94).

## Trigger is razor-thin — ALL THREE required

Remove any one and it passes:
1. a **filleted** body (curved blend faces),
2. a **sphere** subtracted from it (leaves a concave spherical face), then
3. a **cylinder** subtracted from that result (a curved tool).

Verified by these (passing) controls:
- planar chained difference → correct (`997000` vol)
- single curved difference (`cube.fillet − sphere`) → correct
- `(cube − sphere) − cylinder` **without fillet** → correct
- `(cube − cylinder) − cylinder` (no sphere) → correct
- `(cube.fillet − sphere) − smallcube` (box second tool) → correct

## Tracking test

A faithful, currently-`#[ignore]`d regression test is committed at:

```
crates/vcad-kernel/tests/chained_curved_boolean_regression.rs
→ chained_curved_boolean_resurfaces_sphere   (#[ignore])
→ chained_planar_boolean_is_robust           (guard, passing)
→ single_curved_difference_is_robust         (guard, passing)
```

Run it today:
```
cargo test -p vcad-kernel --test chained_curved_boolean_regression -- --ignored
# panics: result z-max = 94.0 (body is ~28; sphere would push to ~94)
[...]

PR #299, opened 2026-06-17
fix(loon): let-bound subjects no longer drop their cuts

ecto:
## What

Fixes the reported bug where multi-feature parts authored via nested / `let`-bound `difference` in `create_cad_loon` lost their cuts. Investigation turned up **two distinct bugs**, not one.

### Bug 1 — loon `let` dropped its body ✅ fixed
`[let name value body]` evaluated the value, bound it, and returned **the bound value** — silently ignoring the trailing body. So `[let b BODY expr]` returned the un-cut `b`, dropping every boolean. (Case 4 produced 2 nodes instead of the 12-node chain.)

The behavioral fix is in **loon-lang's `eval_let`** (`ecto/loon`, consumed by vcad as a path dependency): the 2-arg statement form `[let x v]` is unchanged (still binds into the current scope so top-level / `[do …]` sequencing observes it); `[let x v body…]` now pushes a scope, binds, evaluates the body, and returns its last value.

> ⚠️ **Cross-repo:** the `eval_let` change lives in `ecto/loon` (branch `pr8-fix`), committed there as `41a3112`. It must be pushed/landed in `ecto/loon` for vcad CI/prod (which builds loon via path dep) to pick up the fix. This vcad PR contains the tests + tracking only.

### Bug 2 — kernel chained curved boolean ⚠️ tracked, not fixed → ecto/vcad#298
Case 3's render failure is **not** the loon evaluator — the loon→IR conversion and the `vcad-eval` node evaluator compose nested/chained differences correctly (verified on planar geometry, `997000` vol). The fault is in `vcad-kernel-booleans`: subtracting a **curved tool from a solid that already has a concave curved face** resurfaces the subtracted geometry. Razor-thin trigger requiring all three of: filleted body + sphere cut + cylinder cut. Minimal repro `(filleted_cube − sphere) − cylinder` balloons the bbox z from ~28 to ~94.

Captured as an `#[ignore]`d regression test plus two passing guards, and filed as **#298**.

## Tests
- `vcad-loon/tests/chained_booleans.rs` — IR-structure tests for reported cases 3 & 4 (every cut present; case 4 now matches case 3).
- `vcad-eval/tests/loon_boolean_composition.rs` — end-to-end geometry on clean manifolds (reliable `volume()`): nested / `pipe` / `let`-bound chains all apply every cut.
- `vcad-kernel/tests/chained_curved_boolean_regression.rs` — ignored kernel-bug tracker (#298) + planar/single-curved guards.
- loon-lang: `let_with_body_*` regression tests.

All green (437 loon tests, vcad-loon, vcad-eval, vcad-kernel guards). rustfmt clean; no new clippy beyond the codebase baseline.

## Reviewer notes
- `vcad-loon` added as a **dev-dependency** of `vcad-eval` (no cycle — vcad-loon doesn't depend on vcad-eval) to enable the end-to-end loon→geometry test.
- Changelog entry added (`fix`, user-facing via `create_cad_loon`).
- Case status against the original repro: cases 1 & 2 correct; case 4's evaluator bug fixed; cases 3 & 4 on the exact filleted+sphere+cylinder geometry still render wrong pending the kernel fix (#298). Authoring with a single union-of-tools (`A − (B ∪ C ∪ D)`) is a working workaround.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6926,6 +6926,7 @@ dependencies = [
  "vcad-kernel-sweep",
  "vcad-kernel-tessellate",
  "vcad-kernel-text",
+ "vcad-loon",
 ]
 
 [[package]]
--- a/changelog/entries/2026-06-17-loon-let-body.json
+++ b/changelog/entries/2026-06-17-loon-let-body.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-06-17-loon-let-body",
+  "version": "0.9.4",
+  "date": "2026-06-17",
+  "category": "fix",
+  "title": "Loon let-bound parts no longer drop their cuts",
+  "summary": "[let name value body] in create_cad_loon now evaluates and returns the body; previously the trailing body was silently dropped, so a let-bound base solid rendered un-cut with every boolean lost.",
+  "features": ["loon", "create_cad_loon", "booleans"],
+  "mcpTools": ["create_cad_loon"]
+}
--- a/crates/vcad-eval/Cargo.toml
+++ b/crates/vcad-eval/Cargo.toml
@@ -20,3 +20,6 @@ vcad-kernel-geom.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
+
+[dev-dependencies]
+vcad-loon.workspace = true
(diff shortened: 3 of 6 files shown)


ecto:
**Rust checks are red pending the loon-side fix.** vcad CI clones loon `main` (`git clone … ecto/loon`), which doesn't yet contain the `eval_let` change, so the new `let` tests fail at runtime (clippy + fmt pass — the test code compiles fine). The fix is up as ecto/loon#53; once it merges to loon `main`, re-running CI here will go green. Nothing in this PR needs changing.

PR #306, opened 2026-06-19
feat(mcp): Discord ping on each new MCP session

ecto:
## What

Posts a one-off Discord embed whenever a **new MCP session** is created — so we can see in real time when someone (or some agent) starts using the hosted MCP, alongside the existing 15-minute activity rollup.

Fires from the dispatch layer exactly when a tool mints a `document_id` with **no incoming `document_id`** — i.e. a session *creator* (`open_document`, `create_cad_loon`, `import_step`, `create_schematic`, `sheet_metal_create`, `load_document`). Mutators on existing sessions don't trigger it.

## Behavior

- **Env-driven webhook**, same convention as `packages/app/api/_lib/discord.ts`: `DISCORD_WEBHOOK_URL` (with an optional `DISCORD_WEBHOOK_URL_SESSION` override).
- **Production-gated** (`VERCEL_ENV=production`); set `DISCORD_FORCE=1` to test from preview/dev. No webhook configured → **silent no-op**, so local/stdio runs and the test suite stay offline.
- **Awaited but bounded** by a 2.5s `AbortController` timeout and never throws — a Discord outage can't break or stall a tool call.
- **Privacy**: sends only the session id, the creating tool, a masked `who` (e.g. `a***@example.com` / `anonymous`), and the build sha — never argument values or IR.

## Testing

- 6 new tests: no-op gating (no webhook / non-prod), embed shape + caller masking, session-webhook override, never-throws on POST failure.
- Full `@vcad/mcp` suite: **216 passed**. `tsc --noEmit` clean.
- No changelog entry — internal ops observability, not user-facing product behavior.

## To enable in prod

Set `DISCORD_WEBHOOK_URL` (or `DISCORD_WEBHOOK_URL_SESSION`) on the `vcad-mcp` Vercel project. Until then it's a silent no-op.

### DIFF
--- a/packages/mcp/src/__tests__/notify.test.ts
+++ b/packages/mcp/src/__tests__/notify.test.ts
@@ -6,7 +6,7 @@ import {
   afterEach,
   vi,
 } from "vitest";
-import { fireToolAlert, notifyConfig } from "../notify.js";
+import { fireToolAlert, fireSessionAlert, notifyConfig } from "../notify.js";
 
 const WEBHOOK = "https://discord.com/api/webhooks/test/token";
 // Tests run with a short rollup window so a single timer tick fires it.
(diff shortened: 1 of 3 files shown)


ecto:
Superseded by #305: the new-session Discord ping is better done as a DB trigger that reuses the existing pg_net webhook (vault `discord_webhook_url`, migration 021) — no duplicate webhook config, no Vercel env, and no double-ping with the existing documents-insert trigger. See migration 026 in #305.

PR #379, opened 2026-07-01
Differentiable seam M0–M2: frozen tessellation, lift-bridge, implicit vertex diff

ecto:
First executable piece of the differentiable CAD→physics loop: **dx/dθ**, the sensitivity of tessellated mesh-node positions to a CAD parameter, computed by cutting the chain at the mesh — never differentiating through boolean/trim combinatorics. Design note: [`docs/differentiable-seam-m0-m2.md`](docs/differentiable-seam-m0-m2.md).

## What's in here

- **`vcad_kernel_tessellate::frozen`** — frozen-tessellation mode. `capture_plan` runs the stock tessellator once at the base θ and records per-node recipes (topology vertex, or frozen `(u,v)` surface sample) plus frozen connectivity and an order-invariant topology signature. `evaluate_plan` re-emits `f64` positions against a rebuilt B-rep, recovering entity correspondence *geometrically* (the boolean pipeline enumerates isomorphic output in nondeterministic order — its cap faces empirically swap between `r` and `r−1e-6`). A topology change under perturbation is a hard `TopologyChanged` error; a lost/ambiguous entity match is a hard `CorrespondenceLost` error — never a silent wrong mesh.
- **`vcad-kernel-diff`** (new crate, unwired/WIP like `vcad-kernel-stocksim`) — the seam:
  - the **lift-bridge**: `SurfaceKind` match → concrete struct → existing `lift::<Dual<f64>>()`, with explicit θ→field seeding (`SurfaceSeed`/`ParamSeeding`); inapplicable seeds are hard errors, and `seed_where` returns a count so callers assert exactly which surfaces θ touches. All seven surface kinds have arms. The `Surface` trait, geometry store, and topo crates are untouched.
  - **implicit vertex differentiation** (Pillar 3): each incident surface contributes a row `∇g·ẋ = −∂g/∂θ`; underdetermined tangential DOF are frozen at zero (the frozen-branch convention); dependent rows are consistency-checked and near-parallel gradients are treated as dependent rather than amplifying rhs noise. Incidence is topological ∪ geometric — boolean sewing leaves rim vertices with half-edges on only one of their two defining faces.
  - seam assembly (`SeamMesh`: positions + velocities + frozen triangles), a scalar-generic `mesh_volume` shared with the FD oracle so `dV/dθ` falls out of `Dual` arithmetic, and central-difference oracle helpers.

## Acceptance gates (all ≤ 1e-6 required; measured)

| Milestone | Check | Measured |
|---|---|---|
| M0 extrude `d`, QoI = V | seam dV/dd vs analytic area | exact (0.0) |
| M0 | seam dV/dd vs FD; node-wise dx/dd vs FD | 8.3e-12; 2.9e-11 |
| M1 cylinder radius | node-wise dx/dr vs FD (interior lateral samples exact-radial to 1e-12) | 5.6e-10 |
| M1 | blind→through-hole perturbation | hard `TopologyChanged`, not a lie |
| M2 boolean through-hole | seam dV/dr vs discrete closed form −N·sin(2π/N)·r·t | 1.1e-15 |
| M2 | seam dV/dr vs FD; node-wise incl. rim implicit-diff | 6.2e-10; 5.6e-10 |

M2's *continuum* −2πrt differs from the seam by exactly the rim polygonization factor `sin(x)/x`, `x = 2π/N` (asserted against the `x²/6` bound) — discretization is a property of the mesh, not the seam; the derivative is exact for the object the physics side will consume.

## Notes for review

- The design note records where the spec's ground-truth map needed correction: `TriangleMesh` is `f32` (frozen path emits its own `f64` mesh), the boolean pipeline is order-nondeterministic (signature is a face-descriptor multiset; correspondence is geometric with per-node anchor verification), boolean sewing doesn't twin rim vertices, and primitive cylinder caps are degenerate single-vertex loops.
- Known limitations are documented deliberately (design note §Known limitations + `recipe_rank` docs): recipe priority assumes moving trims live on curved surfaces; geometric incidence uses unbounded implicit forms; tolerances are absolute mm constants.
- Scope fence respected: no reverse mode, no fillets, no generic curves/store, no optimizer. `vcad-kernel-diff` is intentionally not wired into `vcad-kernel`/WASM yet (listed in CLAUDE.md's WIP-crates section). No changelog entry — kernel-internal infra with no user-facing surface.
- Local validation ran `cargo test`/`cargo clippy -- -D warnings`/`cargo fmt --check` on the workspace excluding the loon/phyz-dependent crates (this sandbox cannot clone those siblings; none are touched here). CI runs the full set.

**What M3 needs from this** (short version; full paragraph in the design note): a phyz functional plugs in at `SeamMesh` — contract `∂J/∂x_i` against `velocities`, exactly as `volume_with_derivative` does through `Dual` arithmetic; re-capture per optimizer step; `TopologyChanged`/`CorrespondenceLost` are the subgradient-crossing signals the optimizer must handle.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -111,6 +111,8 @@ vcad/
 │
 │   # Work-in-progress crates (built but not yet wired to any consumer):
 │   # - vcad-kernel-stocksim     # CAM stock sim (octree SDF, marching cubes); no consumers yet
+│   # - vcad-kernel-diff         # Differentiable seam: dx/dθ of frozen tessellations
+│   #                            # (docs/differentiable-seam-m0-m2.md); no consumers yet
 ├── packages/                      # TypeScript workspace
 │   ├── app/                       # Web app (React + Three.js + Zustand)
 │   ├── engine/                    # WASM engine wrapper + physics
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7283,6 +7283,7 @@ dependencies = [
 name = "vcad-kernel-tessellate"
 version = "0.9.4"
 dependencies = [
+ "tang",
  "vcad-kernel-geom",
  "vcad-kernel-math",
  "vcad-kernel-primitives",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7091,6 +7091,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "vcad-kernel-diff"
+version = "0.9.4"
+dependencies = [
+ "tang",
+ "vcad-kernel",
+ "vcad-kernel-geom",
+ "vcad-kernel-math",
+ "vcad-kernel-nurbs",
+ "vcad-kernel-primitives",
+ "vcad-kernel-sketch",
+ "vcad-kernel-tessellate",
+ "vcad-kernel-topo",
+]
+
+[[package]]
 name = "vcad-kernel-drafting"
 version = "0.9.4"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
     "crates/vcad-kernel-step",
     "crates/vcad-kernel",
     "crates/vcad-kernel-constraints",
+    "crates/vcad-kernel-diff",
     "crates/vcad-kernel-wasm",
     "crates/vcad-kernel-drafting",
     "crates/vcad-cli",
@@ -90,6 +91,7 @@ default-members = [
     "crates/vcad-kernel-step",
     "crates/vcad-kernel",
     "crates/vcad-kernel-constraints",
+    "crates/vcad-kernel-diff",
     "crates/vcad-kernel-wasm",
     "crates/vcad-kernel-drafting",
     "crates/vcad-cli",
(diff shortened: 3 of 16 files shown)


ecto:
Re: the three blocking Choji findings on `f3d6923` — I checked each against the source; none require a code change. Details so the review can be dismissed with context:

**1. "TopoVertex anchor check is skipped when the vertex is not in the canonical index"** — the described code path doesn't exist. The actual branch (`seam.rs`) is:

```rust
let vid = *ci
    .vertices
    .get(vertex as usize)
    .ok_or(FrozenError::RecipeOutOfRange)?;   // missing index → hard error
let x = topo.vertices[vid].point;
anchor_check(i, &x, &plan.base_positions[i])?; // unconditional, before any use
```

There is no `ci.vertex_index` lookup, no `None` fallback, and no path that skips `anchor_check`. (There's also no `.position` field — it's `.point`.)

**2. "Cylinder gradient is unnormalized, so the near-parallel threshold scales with radius"** — the suggested fix is already the implemented behavior. `solve_vertex_velocity` normalizes every row before the dependency check:

```rust
let norm = row.gradient.norm();
let mut g = row.gradient / norm;   // unit gradient
let mut r = row.rhs / norm;        // rhs scaled by the same factor
```

`DEPENDENT_TOL` is therefore applied to unit vectors and is radius-independent. `ConstraintRow` deliberately keeps the raw `∇g` (documented as such); normalization is the solver's job so the mathematical form of each implicit surface stays legible.

**3. "Rim-node count `assert_eq!(rim_nodes, 2 * SEGMENTS)` may be wrong"** — the count is verified empirically (the boolean emits exactly `SEGMENTS` rim vertices per cap: 72 topology vertices = 8 block corners + 2×32 rim) and it is deterministic in `segments`, not platform-dependent. The hard equality is deliberate: if the boolean's rim discretization ever changes, the frozen-capture assumptions should be re-examined, and a `>=` bound would hide exactly that signal.

The one actionable finding from the previous round (restricting the geometric incidence scan to face-referenced surfaces) was applied in `f3d6923`.

---

PR #380, opened 2026-07-01
Differentiable seam M3: boundary recipes, mass-property QoIs, and the first geometry grown by gradients

ecto:
Follow-up to #379. M0–M2 proved **dx/dθ** correct against a finite-difference oracle; M3 makes the seam *do* something: the first shape improved by gradient descent through it, plus the two pieces that unlock it. Design note: [`docs/differentiable-seam-m3.md`](docs/differentiable-seam-m3.md).

## 1. Multi-surface boundary recipes — closes M0–M2's known correctness gap

The old recipe-priority heuristic assumed a moving trim always lives on the *curved* surface; the inverse case (e.g. cylinder **height** as θ) froze boundary nodes on the fixed surface — and the FD oracle, replaying the same recipes, agreed with the wrong answer. Now a node contributed by two faces on distinct surfaces becomes `NodeRecipe::Boundary`: position from a Newton solve of `{g_a = 0, g_b = 0, tangent-pinned}`, velocity from the same two-row implicit system as rim topology vertices — the node tracks the moving trim **no matter which surface carries θ**.

- Implicit forms `(g, ∇g)` moved to `vcad-kernel-geom::implicit_form` (one source of truth for tessellate's Newton solve and diff's seeded rows).
- The degenerate-cap tessellation (with and without holes) is factored out of `tessellate_brep` into a shared `tessellate_degenerate_cap` helper used by both the renderer path and frozen capture — the drift-prone duplication flagged in #379's review is gone, and capture now handles boolean-cut caps (previously a hard `UnsupportedFace` error).
- Regression test on the previously-wrong case (`m3_boundary_trim.rs`): rim rides the moving cap at exactly ẑ; `dV/dh` vs closed form and FD at ~1e-10 (gate 1e-6).

## 2. Mass-property QoIs with exact θ-derivatives + the physics hook

`mass_properties` / `mass_properties_with_derivative`: volume, mass, centroid, inertia tensors (origin + centroid) via exact signed-tetra integrals, generic over `tang::Scalar` — `Dual<f64>` positions give every field's derivative in one pass (`Dual<Dual<f64>>` reaches second derivatives). Gates: cuboid closed forms to 1e-12, extrusion `dI_zz/dd` closed form, cylinder-radius FD ~1e-9.

`contract_sensitivity(seam, ∂J/∂x)` is the hook a phyz functional plugs into (`dJ/dθ = Σᵢ ∂J/∂xᵢ · dxᵢ/dθ`); against the analytic volume gradient it reproduces the dual-number derivative to 1e-12. **Phyz-coupling status:** this sandbox can't build the `phyz` sibling (repo scope), so the rollout adapter itself — a small function in `vcad-kernel-physics` — is deferred to a session with phyz access; the interface it targets is frozen and exercised here by analytic physics objectives (inertia-based).

## 3. The optimizer harness and the demo

`objective_gradient` (re-capture per iterate, one seam pass per parameter) + `minimize` (projected gradient descent, warm-started backtracking). Frozen-tessellation errors during *trial* steps — `TopologyChanged`, `CorrespondenceLost`, `BoundarySolveFailed` — are handled as subgradient signals (shrink the step), never silently accepted.

**`m3_flywheel_optimize.rs`:** a disc flywheel with a center bore and four lightening holes (a five-boolean chain per iterate); θ = (bore radius, hole radius); brief = *hit a target spin inertia with minimum mass*. From θ₀ = (3, 2.5) mm, 83 monotone iterations grow the design to (12.0, 4.96) mm — cutting ~13% mass, landing I_zz within 1% of target — with the analytic gradient FD-audited at the start point and several hundred boolean rebuilds under the correspondence machinery. The test asserts monotone descent iterate-by-iterate, convergence (not iteration exhaustion), recovery of most of the achievable objective, and the design brief.

## Validation

- All M0–M2 gates stay green (boundary recipes now also serve the M1 cylinder cap rims and are exercised by the existing tests).
- `cargo test` / `cargo clippy -- -D warnings` / `cargo fmt --check` clean on the workspace excluding the loon/phyz-dependent crates (sandbox limitation, unchanged from #379; CI runs the full set).
- No changelog entry: kernel-internal, still unwired to consumers (CLAUDE.md WIP-crates entry updated).

## What M4 needs from this

Fillet radii put the moving trim on a *tangency* curve, where `∇g_a × ∇g_b` degenerates — M4 is the curvature-aware boundary solve plus torus/cone implicit forms, both localized behind the same recipe and seam interfaces.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -111,8 +111,9 @@ vcad/
 │
 │   # Work-in-progress crates (built but not yet wired to any consumer):
 │   # - vcad-kernel-stocksim     # CAM stock sim (octree SDF, marching cubes); no consumers yet
-│   # - vcad-kernel-diff         # Differentiable seam: dx/dθ of frozen tessellations
-│   #                            # (docs/differentiable-seam-m0-m2.md); no consumers yet
+│   # - vcad-kernel-diff         # Differentiable seam: dx/dθ of frozen tessellations,
+│   #                            # mass-property QoIs, optimizer harness (docs/
+│   #                            # differentiable-seam-m0-m2.md, -m3.md); no consumers yet
 ├── packages/                      # TypeScript workspace
 │   ├── app/                       # Web app (React + Three.js + Zustand)
 │   ├── engine/                    # WASM engine wrapper + physics
--- a/crates/vcad-kernel-diff/src/contract.rs
+++ b/crates/vcad-kernel-diff/src/contract.rs
@@ -1,0 +1,53 @@
+//! The physics hook: contract an external functional's mesh-gradient
+//! against the seam's sensitivities.
+//!
+//! A physics engine (phyz) evaluating a functional `J` on the frozen mesh
+//! supplies `∂J/∂x_i` per node — from an adjoint rollout, analytic
+//! formulas, or its own AD. The shape derivative is then the contraction
+//! `dJ/dθ = Σ_i (∂J/∂x_i) · (dx_i/dθ)` with the seam's
+//! [`SeamMesh::velocities`]. Nothing else about the physics side needs to
+//! know how the geometry was made.
+
+use vcad_kernel_math::{Point3, Vec3};
+
+use crate::SeamMesh;
+
+/// `dJ/dθ = Σ_i (∂J/∂x_i) · (dx_i/dθ)`.
+///
+/// `dj_dx` must be indexed like the seam's nodes (one gradient per node).
+pub fn contract_sensitivity(seam: &SeamMesh, dj_dx: &[Vec3]) -> f64 {
+    assert_eq!(
+        dj_dx.len(),
+        seam.velocities.len(),
+        "∂J/∂x must have one entry per seam node"
+    );
+    dj_dx
+        .iter()
+        .zip(&seam.velocities)
+        .map(|(g, v)| g.dot(*v))
+        .sum()
+}
+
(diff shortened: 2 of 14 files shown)


ecto:
Re: the remaining Choji warning (step-hint growth, `optimize.rs`) — skipping the suggested cap, deliberately.

The hint is set to 2× the **last accepted** step, so the backtracking cost after an overshoot is bounded by `log2` of the step-size *ratio between consecutive iterates*, and the hint re-adapts downward the moment a smaller step is accepted — the "~43 halvings per iterate, every iterate" scenario would require the hint to stay huge while accepted steps stay tiny, which the update rule makes impossible (one doubling per accept). Meanwhile the suggested cap (`initial_step × 4 = 8`) would reintroduce the exact failure the warm-start exists to fix: in the flywheel's flat region the objective needs steps ~100–400× the initial hint (gradient magnitudes ~5e-3), and with a cap at 8 the run stalls at the iteration budget — that's the empirically observed behavior before the uncapped warm-start landed (see the commit history: `MaxIters` at 100 iterations with the capped variant, `StepConverged` at 83 without).

The genuinely tighter-topology concern (a hole approaching the bore) is handled one level up: a trial step that crosses a topology boundary fails with `TopologyChanged`/`CorrespondenceLost` and shrinks — at most a handful of extra rebuilds per such event, then the hint tracks the new, smaller accepted scale.

---

PR #381, opened 2026-07-02
Differentiable seam M4: the fillet radius

ecto:
## What this is

M4 of the differentiable seam (M0–M2: #379, M3: #380): **dV/dr of a fillet radius** — the canonical "real CAD" parameter, and the hard case for a frozen-topology seam. The proving model is the all-edges-filleted cube (6 inset planes, 12 quarter-cylinder edge blends, 8 sphere-octant corners): one θ moves **twenty surfaces at once**, every moving trim is a *tangent* contact, and the fillet kernel rebuilds blend frames nondeterministically between builds.

Design note: `docs/differentiable-seam-m4.md`.

## What landed

**Composite seeds.** A fillet radius perturbs no surface through a single field: each blend's radius grows at rate 1 while its axis simultaneously retreats from the edge. `ParamSeeding` now maps a surface to a *list* of `SurfaceSeed`s, applied additively by the lift-bridge and the implicit rows. The M0–M3 single-seed shape is the singleton case; all existing call sites keep their meaning.

**Frame transport (`FaceFrame` / `transport_uv`).** The fillet pipeline picks blend-frame axis signs by hash order, so a rebuild at θ+h can produce the same geometric surface with a flipped axis — silently changing what a frozen `(u, v)` means and poisoning the FD oracle. The plan now snapshots each face's parameterization frame at capture, and evaluation transports frozen samples through the rebuilt frame (angle/height transport for cylinders and spheres, anchor projection for planes). Face-slot resolution matches on the **worst** transported-sample distance over all of a slot's samples, and every node is still anchor-verified after transport (`CorrespondenceLost` on failure, never a silent wrong answer).

**Tangency completion (`tangency_rows`).** The M3 note predicted the tangent trim would need a second-order boundary solve. It doesn't:

- *On the tangent line*, there is nothing to solve — a frozen-`(u,v)` interior sample with the composite seed already tracks it exactly (retreat `(0,1,−1)` + radial growth `(0,0,1)` compose to the true slide `(0,1,0)`), and the Boundary-upgrade check refuses the singular tangent pair, routing those nodes to the lift-bridge.
- *At corner vertices* where tangent lines cross, the implicit system is genuinely rank-deficient (a surface tangent to a plane contributes a row parallel to the plane's row), so the frozen-direction completion pinned a vertex that actually slides diagonally along its support face. `tangency_rows` adds the missing tangent-curve constraint — first-order rows (`q = axis × n` for a cylinder, two rows spanning `n⊥` for a sphere), no curvature solve, vanishing identically for non-tangent configurations.

Recipe preference during capture also learned curvature rank (plane < singly-curved < doubly-curved) for `(u,v)`-freezing at shared nodes.

## Gates (`m4_fillet_radius.rs`)

| Gate | Requirement | Measured |
|---|---|---|
| Seam dV/dr vs FD oracle (h = 1e-6) | ≤ 1e-6 | **1.7e-8** |
| Node-wise dx/dr, all 836 nodes | ≤ 1e-6 | **3.2e-9** |
| Tangent-line node slide | exact `(0,1,0)` | < 1e-9 |
| Corner-vertex diagonal slide | exact `(1,1,0)` | < 1e-9 |
| Continuum (Minkowski closed form) | within polygonization band | 6.7% (< 10%) |
| Optimizer recovers target radius | — | r* = 2.2 hit to 1e-3 from r₀ = 1.0 |

The optimizer test closes the loop on the fillet parameter itself: projected gradient descent through the seam picks the radius whose rounded cube hits a target volume, reading the radius back from the built model at every iterate so the seeding stays honest.

## Validation

- `cargo test` on the touched crates + scoped workspace suite: **1604 passed, 0 failed**
- `cargo clippy --all-targets -- -D warnings` clean on touched crates
- `cargo fmt --all --check` clean

## Notes

- The model uses `fillet_all_edges`; single-edge `fillet_edges_detailed` currently produces a malformed solid (insets all six faces, emits one blend) — a pre-existing fillet-kernel bug independent of the seam, which the seam's signature/anchor checks are exactly the machinery that caught. Filed in the design note.
- No changelog entry: `vcad-kernel-diff` remains a WIP crate with no consumers (per the changelog policy).

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,8 +112,9 @@ vcad/
 │   # Work-in-progress crates (built but not yet wired to any consumer):
 │   # - vcad-kernel-stocksim     # CAM stock sim (octree SDF, marching cubes); no consumers yet
 │   # - vcad-kernel-diff         # Differentiable seam: dx/dθ of frozen tessellations,
-│   #                            # mass-property QoIs, optimizer harness (docs/
-│   #                            # differentiable-seam-m0-m2.md, -m3.md); no consumers yet
+│   #                            # mass-property QoIs, optimizer harness, differentiable
+│   #                            # fillet radius (docs/differentiable-seam-m0-m2.md,
+│   #                            # -m3.md, -m4.md); no consumers yet
 ├── packages/                      # TypeScript workspace
 │   ├── app/                       # Web app (React + Three.js + Zustand)
 │   ├── engine/                    # WASM engine wrapper + physics
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7096,6 +7096,7 @@ version = "0.9.4"
 dependencies = [
  "tang",
  "vcad-kernel",
+ "vcad-kernel-fillet",
  "vcad-kernel-geom",
  "vcad-kernel-math",
  "vcad-kernel-nurbs",
--- a/crates/vcad-kernel-diff/Cargo.toml
+++ b/crates/vcad-kernel-diff/Cargo.toml
@@ -16,5 +16,6 @@ vcad-kernel-primitives = { workspace = true }
 vcad-kernel-tessellate = { workspace = true }
 
 [dev-dependencies]
+vcad-kernel-fillet = { workspace = true }
 vcad-kernel-sketch = { workspace = true }
 vcad-kernel = { workspace = true }
--- a/crates/vcad-kernel-diff/src/implicit.rs
+++ b/crates/vcad-kernel-diff/src/implicit.rs
@@ -30,7 +30,9 @@ pub struct ConstraintRow {
     pub rhs: f64,
 }
 
-/// The implicit form of a surface at `x`: `(g, ∇g, ∂g/∂θ)`.
+/// The implicit form of a surface at `x`: `(g, ∇g, ∂g/∂θ)`, with the
+/// θ-term summed over composed seeds (a fillet blend's radius and axis
+/// position both carry the fillet-radius parameter).
 ///
 /// Single source of truth for both [`constraint_row`] and
 /// [`surface_residual`], so incidence detection and the rows actually
@@ -39,7 +41,7 @@ pub struct ConstraintRow {
 /// `Ok(None)` for kinds without an implicit form.
 fn implicit_terms(
     surface: &dyn Surface,
-    seed: Option<SurfaceSeed>,
+    seeds: &[SurfaceSeed],
     x: &Point3,
 ) -> Result<Option<(f64, Vec3, f64)>, DiffError> {
     let kind = surface.surface_type();
@@ -48,11 +50,13 @@ fn implicit_terms(
             let plane = downcast::<Plane>(surface, kind)?;
             let n = *plane.normal_dir.as_ref();
             let g = plane.signed_distance(x);
-            let g_theta = match seed {
-                None => 0.0,
-                Some(SurfaceSeed::Translate { velocity }) => -n.dot(velocity),
-                Some(other) => return Err(DiffError::UnsupportedSeed { kind, seed: other }),
-            };
+            let mut g_theta = 0.0;
+            for seed in seeds {
+                g_theta += match *seed {
+                    SurfaceSeed::Translate { velocity } => -n.dot(velocity),
+                    other => return Err(DiffError::UnsupportedSeed { kind, seed: other }),
+                };
+            }
             Ok(Some((g, n, g_theta)))
         }
         SurfaceKind::Cylinder => {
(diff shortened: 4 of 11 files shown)


ecto:
Responding to the Choji findings (addressed in `accf35a` where action was warranted):

**Major 1 (cylinder transport negates `v` under an axis flip) — not a bug; the negation is the correct transport.** The current code

```rust
let v2 = (*center - c.center).dot(axis2) + uv.y * axis.dot(axis2);
```

*is* the projection identity the finding asks for. The capture-time material point is `p = center + axis·v + r·d` with `d ⊥ axis`, so its height in the rebuilt frame is

```
(p − c.center)·axis2 = (center − c.center)·axis2 + v·(axis·axis2) + r·(d·axis2)
```

and the radial term vanishes because the rebuilt cylinder is the same geometric surface (`axis2 = ±axis ⇒ d·axis2 = 0`). The suggested fix `(p3 − c.center).dot(axis2)` with `p3 = center + axis·uv.y + d·r` expands to exactly the expression already in the code. When `axis2 = −axis`, the same material point genuinely has coordinate `−v` in the flipped frame — evaluation then computes `c.center + axis2·v2 + …`, landing on the identical 3-D point. Transporting `uv.y` unchanged (the finding's stated intent) is what would negate the height.

The "both sides see the same bug" theory doesn't hold either: every transported node is verified against its capture-time anchor within `MATCH_TOL = 1e-3 mm` and errors with `CorrespondenceLost` on failure. A wrongly-signed height moves a blend sample by `2|v|` — millimeters — and trips that check immediately. That failure mode is exactly what motivated `transport_uv`: the fillet kernel really does flip axis signs between builds in the M4 test, and the run only passes because the negation is applied when needed.

**Major 2 (tangency rows over-generated for every plane/surface pair)** — the pair loop over-generates by design and safely: `tangency_rows` returns *no rows* for transverse contacts (the parallelism check `radial × n` gates emission), and `solve_vertex_velocity` is not a least-squares solver — it orthogonalizes rows with the rhs carried, checks any dependent row's rhs for consistency (`InconsistentConstraints` is a hard error), and only then discards it. Redundant rows are harmless by construction; disagreeing ones can never be silently averaged. Added a comment at the loop in `seam.rs` spelling this out, as suggested.

**Minor (`seed_where` composes rather than replaces)** — legitimate doc gap; the doc now states explicitly that it composes with existing seeds, like `seed`, and never overwrites. (For context, no in-tree caller relied on overwrite semantics — the tests assert seeded-surface *counts*, and composition is the entire point of the M4 change.)

The sandbox `npm run test`/`typecheck` failures (exit 127, turbo not found) are Choji-environment issues on a Rust-only diff, per its own notice — no action taken.

---

ecto:
On the two `accf35a` findings:

**Cross-product order in cylinder transport** — verified against the kernel: `transport_uv` and the surfaces use the *same* convention. `CylinderSurface::y_dir()` (and `SphereSurface::y_dir()`) in `vcad-kernel-geom` are

```rust
/// Compute the y direction (perpendicular to axis and ref_dir).
pub fn y_dir(&self) -> tang::Vec3<S> {
    self.axis.as_ref().cross(self.ref_dir.as_ref())
}
```

with `evaluate` reconstructing `ref_dir·cos u + y·sin u` — exactly the expression in `transport_uv`, on both the capture and rebuilt sides. The suggested unit test was worth having regardless, so `d33a56d` adds `transport_uv_survives_reframed_cylinder_and_sphere`: it transports frozen `(u, v)` through a cylinder rebuilt with a flipped axis, a `ref_dir` rotated 0.7 rad, and a center slid along the axis (all three of which the fillet kernel actually does between builds), plus a sphere under a wholly different orthonormal frame, and asserts the rebuilt surface's `evaluate` lands on the same material point to 1e-12. That pins the convention against future drift in either crate.

**Plane-plane pairs in the tangency loop** — confirmed harmless and cheap in the current code: `tangency_rows` matches on `surface_type()` and a plane falls straight through to the wildcard arm

```rust
_ => Ok(Vec::new()),
```

so no downcast ever runs for a plane (the downcasts live inside the `Cylinder`/`Sphere` arms). Declining the suggested pre-filter: `tangency_rows` owning the "which kinds participate" decision means a future surface kind gaining tangency support (cone-on-plane, torus-on-plane) needs exactly one edit, whereas a caller-side kind check would have to be kept in sync by hand. The cost of a plane-plane pair is one enum match per pair per vertex.

---

ecto:
Both doc requests landed in `7909c57`: `transport_uv` now cites the `y_dir = axis × ref_dir` convention it shares with `CylinderSurface`/`SphereSurface`, and `tangency_rows` documents that unsupported kinds return no rows, never an error — callers pass every incident surface and the function decides which pairs participate.

One correction to the remaining concern: the new unit test does *not* only exercise the transport function. It round-trips through the geom crate's own `evaluate` on both sides (`p = cyl.evaluate(uv)` → transport → `assert cyl2.evaluate(uv2) == p`, with a rotated `ref_dir` and flipped axis), so a future refactor of `y_dir()`'s cross-product order in `vcad-kernel-geom` fails this test loudly — the cross-crate consistency is asserted empirically, not assumed. The doc comment now says so explicitly.

---

PR #387, opened 2026-07-02
feat(gdsii): GDSII import/export, hierarchy flatten, and vcad-ir bridge

ecto:
## What

New `vcad-gdsii` crate — the foundation for chip-layout support in vcad:

- **Record-level codec** for the GDSII (Calma) stream format: all 26 core record types, big-endian, NUL-padded strings, unknown records tolerated on read. The 8-byte excess-64 IBM float (UNITS/MAG/ANGLE) round-trips f64 **exactly** — tests assert `==`, not approx.
- **Plain-data model**: `Library → Cell → Element::{Boundary, Path, Text, Sref, Aref}`, i32 DB-unit coords.
- **`flatten()`**: recursive SREF/AREF resolution (translate / rotate / mirror / magnification via 2D affine composition, exact trig for 90° multiples), AREF lattice expansion, reference-cycle detection, PATH→boundary expansion (pathtype 0).
- **`to_vcad_document()`** (`vcad-ir` feature, default on): flattened layers → sketch+extrude parts from a caller-supplied physical film stack `(layer, z_bottom_um, thickness_um, name)`, with a view-scale param (default 1 µm → 1 mm).

## Why

GDSII is *the* interchange format for IC layout. This lets vcad read and write real chip masks — first consumer is rendering dies in the viewport (see `examples/haiku_die.rs`: builds a small NMOS-style die, GDS write → read → flatten → .vcad, renders correctly through vcad-render).

## Testing

- 41 tests (40 unit + 1 doc): exact real8 round-trips, write→read structural equality (SREF w/ rotation+mirror, PATH, AREF 2×3), exact flatten coords for 90°/mirror/mag/nested transforms, miter geometry, cycle detection, golden two-layer count test, bridge node counts/scaling/serialization.
- `cargo clippy -p vcad-gdsii --all-targets -- -D warnings` clean (with and without the `vcad-ir` feature); `cargo fmt` clean.

## Scope notes

- Pathtype 1/2 (round/extended ends) parse but are rejected by the flattener (documented).
- TEXT elements round-trip but aren't mask geometry, so flatten ignores them.
- Datatypes merge per layer in flatten output (documented).


ecto:
**Validated against real foundry GDS.** Ran a small design (loon-hdl `haiku_gen`, 142 FFs / ~800 logic cells) through librelane's full sky130 RTL-to-GDS flow, then imported the final 1.27 MB die (127×127 µm, 1,930 instances, 0 DRC errors) with this crate:

- Parse + flatten: clean on first try — 104,663 polygons across 17 sky130 layers, no unsupported records, no pathtype surprises.
- Found and fixed one real bug in the process (89e8ed88): the IR bridge was folding each layer's extrudes into a left-nested `Union` chain — 38k deep on the li1 layer, which overflowed the stack of the recursive document evaluator in vcad-render. Now unions pairwise into a balanced tree (depth ~log₂ n) with a 1000-instance regression test.
- `examples/sky130_import.rs` is the harness used (auto-detects the top cell, prints per-layer stats, emits a .vcad with a sky130 film stack).

PR #402, opened 2026-07-07
feat(examples): laser-cut glockenspiel — the SCS closed-loop demo part

ecto:
The CAD half of [the SCS closed-loop demo](https://github.com/ecto/vcad/pull/397): a C6–C7 glockenspiel where **every dimension is physics** and every receipt claim has a household oracle — the loudest being a phone spectrogram.

## What's in it

- **`geometry.mjs`** — single source of truth. Bar lengths are *solved* from target pitches via the closed-form free-free beam model `f₁ = 3.5608·(t/L²)·sqrt(E/12ρ)`; E and ρ come from `vcad-kernel-sheet`'s materials registry (`al-hard` = 6061-T6) and are asserted against the live engine at build time. Cord holes sit on the fundamental's nodal lines (0.2242·L from each end). The stand's 16 deck holes sit directly under each bar's nodal holes — two converging rows, the tuning physics visible in the frame.
- **`build.mjs`** — the full pipeline in-process: create → `sheet_metal_check` vs the SendCutSend catalog → unfold → per-part DXF → springback-compensated bend sequence → folded STEP + GLB → cost → `receipt.json`.
- **`frequencies.mjs`** — the tuning table standalone (no build needed): note, target Hz, as-modeled length, predicted Hz, error in cents (≤1.2¢ from rounding to the 0.1 mm cut grid).

## The DFM loop, exercised honestly

The stand's chamfered deck corners put material at the wall-bend ends: the naive fold fails `sheet_metal_check` with 4 tear-out warnings. `sheet_metal_suggest_fix` answers `add_bend_relief`; the build applies it **in the design** (parametric notches, visible in the DXF) and re-checks to zero. The build *asserts* the naive design fails and the fixed one passes — the self-heal loop is part of the test.

Material split is deliberate: SCS cuts 6061 but doesn't bend it (bars: cut-only, raw — coatings damp the ring); the stand uses their 0.125″ 5052-H32 bending stock at its fixed 3.18 mm radius and published K-factor (0.44).

## Verification

- All 9 parts check shop-ready against the `sendcutsend` catalog (0 errors, 0 warnings).
- Stand DXF audited entity-by-entity: 1 closed exterior, all 16 hole loops at the exact nodal-row centers, 4 dashed bend centerlines on the allowance midline.
- Folded STEP is ~12 MB (24-gon cord holes keep the B-rep writer's face count sane; 0.018 mm from true circles — an order of magnitude inside cut tolerance).
- Bars: 3 loops each (outline + 2 nodal holes), total 715 g across the set.

Depends editorially on #397 for the plan doc link; no code dependency.


ecto:
**Added: `simulate_strike` — the audio oracle, closed before the order.** (ebd1a8c4)

New MCP tool that strikes a flat bar in simulation and verifies its pitch through the audio path: hole-aware 1-D FEM modal analysis → mallet-excited synthesis → WAV → FFT peak → cents verdict.

The FEM immediately caught a real detune: the nodal cord holes flatten every bar **≈ −5 cents** (stiffness loss where mode-1 curvature is nonzero) — bigger than the cut-grid rounding error. Cut lengths are now hole-compensated (~0.2 mm shorter), and `build.mjs` gained an order gate: every bar's synthesized strike must FFT within ±5 ¢ of its note or the build fails. All 8 land within ±1.7 ¢; `out/` now ships a WAV per bar you can play against the physical strike when the box arrives.

Timbre physics is in there too: non-harmonic partials at 2.76/5.40/8.93 × f₁, center strike suppressing the antisymmetric mode 2, and cord-at-nodes suspension damping making the partials die in tenths of a second while the fundamental rings for seconds. Frequencies are first-principles; decay Q is a documented heuristic.

The `modes` table `{hz, gain, q}` in the tool result is the data contract for client-side Web Audio synthesis in the viewer widget (tap a bar in chat, hear the note) — follow-up, pending the single-canvas viewer deploy.

12 new tests (439 total pass): exact free-free eigenvalue roots, FEM ≡ closed form on a uniform bar to <0.5 ¢, synth→FFT round trip <1 ¢, strike-position physics, session-doc geometry extraction, bent-part rejection.

PR #469, opened 2026-07-08
Boolean kernel: conforming seams — T-junction repair + canonical arc discretization

ecto:
## What

Attacks priority 1 from the torr retest handoff (2026-07-08): **any intersecting union against a curved or boolean-result mesh leaves unstitched seams**, and any further boolean on the damaged mesh degrades or times out.

Two root causes fixed:

1. **Non-conforming topology (T-junctions).** Trimming split one face's boundary at intersection points while the neighboring face sharing that edge kept it whole — even `cube ∪ cube` left 36 open loop-edges, invisibly poisoning every downstream boolean. New `repair::split_edges_at_interior_vertices` splits unpaired half-edges at existing vertices lying in their interior, reusing the `VertexId` so the pieces pair by id.

2. **Phase-mismatched arc sampling.** Every splitter discretized circles in its own frame at relative fractions of the arc span, so two faces bordering the same circle never emitted the same points. New `split::canonical_arc_points` samples on an absolute angular grid in a frame derived purely from the circle geometry; the planar by-arc splitter and the cylindrical by-line splitter (which now emits dense two-chain loops instead of single-chord rectangles) share it.

## Results

- `cube ∪ cube` / `cube − cube`: fully watertight (were 36 / 18 open edges)
- blade ∪ cylinder: ~590 → ~30 open edges at native resolution, volumes exact
- trim-inside-pattern (A1 sliver family): now analytically exact (427.3 vs 427.3)
- A2 boolean-over-pattern residual: +4.0% → +2.1%
- New `seam_conformity` regression test asserts conforming topology for planar unions
- New F-cluster tests in the torr catalogue capture the whole intersecting-union family; still-open cases are `#[ignore]`d with root-cause notes

## What remains (staged, not in this PR)

The full fix — freezing analytic circle boundaries into canonical polylines at boolean entry so results conform at **any** display resolution — is prototyped on `claude/kernel-seam-freeze-wip` (single cylinders/annuli fully watertight + resolution-independent there, but chained differences and classification interactions still regress; needs its own session).

`cargo test --workspace`, clippy (`--exclude vcad-desktop`), and fmt all clean.


ecto:
Addressed the review + CI failures in 96c81fdc-follow-up commits:

- **Clippy failures (Rust stable/nightly)**: needless-borrow / manual-clamp / unused-mut lints in the new splitter+repair code — fixed; CI-equivalent clippy (`--features vcad-kernel-text/no-builtin-font -D warnings`) now clean locally.
- **TypeScript failure**: `@vcad/engine` pinned an exact triangle count (44) for the corner-cylinder difference; the T-junction repair + canonical arc grid legitimately densify that mesh (44 → 112, shape unchanged, bbox assertions unaffected). Expectation updated with a comment documenting the kernel change, and the checked-in kernel-wasm artifacts rebuilt.
- **Choji finding 1 (`arc_segments` unused)**: it is used — `canonical_arc_points` calls it to pick the grid count — so no dead-code warning; kept as the single shared count rule with the freeze-branch note.
- **Choji finding 2 (3% commutation tolerance)**: took the suggestion — trim-in-pattern is now additionally pinned to its analytic value (8 × 53.4138 at 0.5%), so the widened equivalence tolerance can't hide a regression on the exact side.
- **Choji finding 3 (annulus test asserts nothing)**: intentional-informational, but noted — it documents the Reversed-flag directed-census artifact; the asserting regression test is `cube_union_cube_conformity` in the same file.

PR #470, opened 2026-07-08
fix(sheet-metal): watertight tessellation for flanges with holes

ecto:
## What

Sheet-metal panel meshes for `SheetMetalBaseFlangePolygon` ignored interior hole loops and wound the lateral walls inward, so `inspect_cad` / `verify_spec` measured garbage geometry even though the underlying solid was correct.

**Field repro:** a 64-gon D58 disc flange (t=2.7) with a 16-gon D8.4 bore + four 12-gon D3.3 holes at BCD 22 measured **2374 mm³ vs ~6890 analytic**, watertight **false** (256 unpaired directed edges). Independent evidence the solid was fine: `sheet_metal_cost` reported the correct mass (0.0541 kg = 6870 mm³ × 7.87 g/cc) and the STEP export was dimensionally sound — both read the solid, not this mesh path.

## Root cause

`tessellate_model` in `crates/vcad-kernel-wasm/src/sheet_metal.rs` is the only mesh path for sheet-metal on the web/MCP surface (STEP uses `folded_sheet_solid`). Two defects:

1. **Caps ignored `panel.holes`** and fan-triangulated the outline only (also wrong for concave outlines like L-brackets).
2. **Side-wall quads were wound inward** — for *every* flange ever meshed, not just ones with holes. It never showed visually because the normal *attribute* was computed separately (`edge × n`, correct) and the material renders double-sided; only the signed-volume integral and the directed-edge watertight check caught it. Decomposing the repro's signed volume by face group made it obvious: top fan +1187, bottom fan +1187, sides **−4748** → −2374 mm³.

## The fix

- New public `triangulate_polygon_2d` in `vcad-kernel-tessellate` (earcut wrapper, output normalized CCW by total signed area).
- `tessellate_model` now cuts hole loops into both caps, emits a wall per hole bore, normalizes ring windings locally (outline→CCW, holes→CW; robust to mis-wound input), and winds all walls outward. Model / unfold / DXF conventions untouched.
- **verify_spec `expected: None`:** the receipt schema names the spec-bound field `predicted`, so anything reading claims for "expected" found nothing. Fixed as a display alias — the tool output mirrors `predicted` into an `expected` field; schema and Rust round-trip unchanged.

Repro now measures **6888.14 mm³ vs analytic 6888.1, watertight true**, and `verify_spec` rolls up to `pass`.

## Tests

- `crates/vcad-kernel-wasm/tests/repro_sheet_holes.rs`: repro disc, single-hole square, concave L-shape, mis-wound rings — each asserting watertight + volume within 1% of analytic.
- `packages/mcp/.../verify-spec.test.ts`: both repro geometries grade `pass` on volume/watertight through the real tool, plus an `expected`-alias assertion.
- Full suites green: MCP 698, engine 96, tessellate/wasm/kernel/eval crates all pass. `cargo clippy --workspace -D warnings` and `cargo fmt --check` clean.

## Reviewer notes

- The checked-in `packages/kernel-wasm/vcad_kernel_wasm*` artifacts are **deliberately not touched** (single-writer `wasm-refresh.yml` on main; CI builds WASM from source). I rebuilt them locally only to run the TS tests.
- The `expected` alias is display-only to avoid a schema bump; `predicted` is consumed by the PCB/mech/EM receipt adapters and stays the canonical field.

### DIFF
--- a/changelog/entries/2026-07-08-sheet-metal-hole-tessellation.json
+++ b/changelog/entries/2026-07-08-sheet-metal-hole-tessellation.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-08-sheet-metal-hole-tessellation",
+  "version": "0.9.4",
+  "date": "2026-07-08",
+  "category": "fix",
+  "title": "Sheet-metal flanges with holes tessellate watertight",
+  "summary": "Panel meshes now cut hole loops into the caps, wall each bore, and wind outward — inspect_cad/verify_spec volume and watertightness match the solid instead of reporting garbage.",
+  "features": ["sheet-metal", "mesh", "verification"],
+  "mcpTools": ["sheet_metal_create", "inspect_cad", "verify_spec"]
+}
--- a/crates/vcad-kernel-tessellate/src/lib.rs
+++ b/crates/vcad-kernel-tessellate/src/lib.rs
@@ -1451,13 +1451,20 @@ pub fn triangulate_polygon_2d(
         return None;
     }
     // Normalize the output winding to CCW via the TOTAL signed area of the
-    // triangulation — immune to any individual degenerate triangle.
+    // triangulation — immune to any individual degenerate triangle. A zero
+    // (or non-finite) total means the input was fully degenerate (collinear
+    // points earcut still triangulated into non-empty but zero-area output):
+    // the winding is undefined, so reject rather than return a coin-flip
+    // orientation the caller would pair with mis-wound walls.
     let mut area2 = 0.0;
     for t in tris.chunks(3) {
         let (ax, ay) = (data[2 * t[0]], data[2 * t[0] + 1]);
         let (bx, by) = (data[2 * t[1]], data[2 * t[1] + 1]);
         let (cx, cy) = (data[2 * t[2]], data[2 * t[2] + 1]);
         area2 += (bx - ax) * (cy - ay) - (by - ay) * (cx - ax);
+    }
+    if !area2.is_finite() || area2 == 0.0 {
+        return None;
     }
     let flip = area2 < 0.0;
     Some(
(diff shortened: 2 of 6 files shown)


ecto:
Thanks for the review. Addressing the findings:

- **Major (hole bore walls emitted?)** — Yes. The wall-emission loop iterates over `rings`, which is `once(outline).chain(holes)` — the outline perimeter **and** one bore per hole, all wound outward via `edge × n` (outline CCW / holes CW). That's what makes the holed panels watertight; the four regression tests (repro disc, single-hole square, mis-wound rings) assert `open_directed_edges == 0`, which would fail immediately if any bore wall were missing. No change needed — just was in the truncated diff section.

- **Minor (`area2 == 0.0` degenerate winding)** — Fixed in a7d5448. `triangulate_polygon_2d` now returns `None` when the total signed area is zero or non-finite (fully collinear input that earcut still emits as non-empty zero-area triangles), so the sheet-metal caller skips the (nonexistent) cap rather than pairing a coin-flip orientation with the walls. Added unit tests for the CCW square, hole-cutting, and degenerate-rejection cases.

- **Minor (code ownership)** — Noted; the sole prior owner of `sheet_metal.rs` is @claude, so there's no separate human owner to loop in.

ecto:
Thanks for the owner review. Agreed on the forward-looking note: the `None` path is correct for genuinely zero-area panels (no cap to miss), and an earcut failure on an *otherwise-valid* ring would currently mesh as an open tube on this path rather than falling back to bridge/ear-clip like `earcut_polygon_with_holes` does. Leaving that as-is for this PR since it's non-blocking and the downstream watertight/isoperimetric integrity checks would flag any such panel rather than shipping it silently — but noted as the natural next hardening step if that case ever shows up in practice.

PR #524, opened 2026-07-11
Verified design loop, Acts I–III: claim basis, trust boundary, predict_physics, macro library

ecto:
First three moves of the verified-design-loop plan, in dependency order.

## 1. Receipt claim basis (`vcad-receipt`)

- New `ClaimBasis` enum: `predicted | verified | measured`. Optional on `ReceiptClaim`; **absent = verified** so every existing receipt keeps its meaning (wire schema stays `vcad.receipt/1`).
- New `ReceiptVerdict` rollup with `provisional`: a receipt that would pass but has any claim resting on a surrogate prediction reads **provisional, never pass**. Fail-closed default (`unverifiable`).
- `ReceiptSummary` gains `predicted_basis` count + basis-aware `verdict`; TS producer mirrored; IR types regenerated.

## 2. Commerce trust boundary (`@vcad/mcp`)

Mechanical injection confinement at the single dispatch choke-point, before any handler runs ([docs/trust-boundary.md](docs/trust-boundary.md)):

- **Opaque ids only** on the money plane (`order_id`/`authorization_id`/`idempotency_key`/`document_id`) — a "part number" from a poisoned datasheet can never be an order argument.
- **Store-scoped artifact refs**: bare `art_…` ids, `/artifacts/…` paths (dot-segment traversal refused), or vcad-host URLs only.
- **Plain fab-bound text**: `ship_to`/`material`/`finish` refuse URLs, control characters, unbounded fields.
- Fail-closed `TRUST_BOUNDARY:` refusals; smuggling tests are the CI proof.

## 3. `predict_physics` — two-tier static FEA (the fast inner loop)

- Kernel: standalone `analyze` module on the existing topopt voxel-hex FEA (`vcad-kernel-topopt/src/analyze.rs`) — max displacement, von Mises stress, compliance. Unit-E solve rescaled by the real modulus; validated against Euler–Bernoulli beam theory.
- **Resolution is the fidelity dial, same solver both tiers**: `fidelity=predict` (res 32, ~100 ms) stamps claims `basis=predicted` → receipt reads **provisional**; `fidelity=verify` (res 72) stamps `basis=verified` → certifiable pass. This is the honest two-tier pattern #1 exists to support.
- Predict-tier floor is 32: trilinear hexes lock in bending below ~4 elements through the thinnest section (res-20 cantilever reads 2.2× too stiff; documented in code + tests).
- WASM `analyzeStaticsBox`/`analyzeStaticsMesh` + engine wrappers; MCP tool is read-only (box runs touch no session).

## Verification

- Rust: `vcad-receipt` 24, `vcad-kernel-sheet` 131, `vcad-kernel-topopt` 23+1 (release) — clippy `-D warnings` (own crates), fmt, `ir:check` all clean
- TS: full `@vcad/mcp` suite **795 passed** (incl. 7 smuggling tests + 5 predict_physics tests exercising the WASM path end-to-end); full `tsc --noEmit` clean
- Tool-surface fixture regenerated for the new tool (deliberate surface change)
- WASM artifacts reverted to origin/main per policy (wasm-refresh.yml is the single writer); Cargo.lock phyz drift dropped


ecto:
**Act III added: the agent macro library** (`define_loon` / `call_loon` / `list_loons` + `use_loons` on `create_cad_loon`).

A macro is named loon source — `[let <name> [fn [params…] …]]`, the exact idiom the stdlib is written in — prepended to programs the same way the stdlib is, so this needed **zero engine or language changes**. The trust ladder starts at definition time: `define_loon` refuses source that doesn't compile or whose example-argument smoke call doesn't yield geometry, so only known-good macros enter the library. `call_loon` instantiates by positional args; `use_loons` composes stored macros inside arbitrary programs (tested: two flange instances united in one scene).

Storage v1 is process-warm + local JSON files under `VCAD_MCP_STATE_DIR/loon-macros`; hosted per-user durability (`mcp_macros` table + migration) and receipt-certified macros are the next rung.

Suite: **804 passed** (9 new macro tests). WASM artifacts again reverted to origin/main.

PR #525, opened 2026-07-11
Persist predict_physics claims: physics_specs + receipt re-verification

ecto:
Mirrors the clearance-claim persistence pattern (check_clearance) for `predict_physics`, so structural limits re-verify via `build_receipt` / `verify_receipt` instead of rotting as one-off checks.

**Stacked on #524** (needs ClaimBasis + predict_physics). Base will be retargeted to `main` once #524 merges.

## What

- **IR**: `PhysicsSpec` (label, part/domain_box, loads, supports, material props, limits, fidelity) + `Document.physics_specs`, ts-rs exported (`npm run ir:gen` regenerated; `gen-ir-types.mjs` marks the field wire-optional like `clearance_specs`).
- **predict_physics `label` arg**: upserts the spec on the document; the part is persisted by resolved root id so it survives renames. Requires `document_id` and at least one limit.
- **`physicsReceiptClaims`**: re-solves each spec at its *stored* fidelity, emitting `physics.static.<label>.{displacement,stress}` claims with `basis: predicted|verified` (ClaimBasis from #524) and a re-runnable `StoredPhysicsClaim` payload in `details`.
- **`verifyPhysicsClaims`**: re-solves per claim and classifies **Holds** (identical result), **Stale** (geometry changed, limit still holds), **Violated** (limit exceeded). One solve per spec via cache when a spec carries both limits.
- **Fail-closed**: unresolvable part, load region off the structure, non-converged solve, or a tampered/missing details payload → unverifiable claim at build time, Violated at verify time. Never a silent pass.
- **Wiring**: `build_receipt` emits physics claims in both the mechanical-only and PCB branches; `verify_receipt` re-verifies them and folds into the worst-status rollup. Drive-by fix: the `verify_receipt` tool handler never passed the engine, so clearance re-verification via the live tool always errored — now passes `ctx.engine`.

## Tests

- New `physics-claims.test.ts`: persistence/upsert, labeled claim emission with basis, Holds/Stale/Violated transitions, fail-closed paths (10 tests).
- Full `@vcad/mcp` suite: 805 passed. `cargo test -p vcad-ir`, workspace clippy `-D warnings` clean. Tool-surface fixture regenerated.
- No WASM artifacts committed (rebuilt locally for tests only).

### DIFF
--- a/changelog/entries/2026-07-11-physics-spec-persistence.json
+++ b/changelog/entries/2026-07-11-physics-spec-persistence.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-11-physics-spec-persistence",
+  "version": "0.9.4",
+  "date": "2026-07-11",
+  "category": "feat",
+  "title": "Persisted physics assertions re-verify via receipts",
+  "summary": "predict_physics gains a label arg: the spec persists on the document, build_receipt re-solves it as physics.static claims, and verify_receipt classifies Holds/Stale/Violated.",
+  "features": ["physics", "receipt", "verification"],
+  "mcpTools": ["predict_physics", "build_receipt", "verify_receipt"]
+}
--- a/crates/vcad-ir/src/lib.rs
+++ b/crates/vcad-ir/src/lib.rs
@@ -1807,6 +1807,10 @@ pub struct Document {
     /// `check_clearance` and receipt verification whenever geometry changes.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub clearance_specs: Vec<ClearanceSpec>,
+    /// Named static-physics assertions (loads, supports, limits), re-solved by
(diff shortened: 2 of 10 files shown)


ecto:
Added e0eec02f: the place_order M4 receipt gate now also re-solves persisted physics specs — failing/unverifiable physics.static claims refuse the debit fail-closed, and the no-specs 'unverified' path requires neither clearance nor physics specs. Two new gate tests; full mcp suite 807 passed.

ecto:
Pushed 215fcb8f: the two new order-gate tests timed out at vitest's 5 s default on CI (two mesh FEA solves each) — they now solve at resolution 12 with a 30 s allowance. Also took Choji's epsilon finding: STALE_REL_EPS raised 1e-9 → 1e-6 with a comment explaining that PCG converges to a tolerance, not bit-identical results.

PR #537, opened 2026-07-15
feat(vcad-render): clap CLI with -o output and batch mode

ecto:
## What

Modernizes the `vcad-render` CLI ([crates/vcad-render/src/main.rs](../blob/claude/vcad-render-improvements-c61478/crates/vcad-render/src/main.rs)), replacing the hand-rolled argument parser with **clap derive**.

- **`-o/--output <path>`** — format inferred from the extension (`.svg`/`.jpg`/`.jpeg`); `-o -` writes SVG to stdout.
- **Batch mode** — multiple input paths or a directory (expands to sorted `*.vcad`), each rendered to `<stem>.<ext>` next to the input or in `--out-dir`, with `--format svg|jpeg`. A per-file failure is reported on stderr without aborting; the batch exits `2` if any file failed.
- **`--help` / `--version`** now work properly.

## Backward compatibility

Every existing flag (`--view`, `--scale`, `--transparent`, `--jpeg`, `--size`, `--fill`, `--quality`) and the positional input path are preserved. `--jpeg <path>` remains as a legacy alias for `-o <path.jpg>`. The exact mecheval invocation shape (`<file> --view front --jpeg out --size --fill --quality`) is unchanged, so `mecheval/scripts/render-task-assets.mjs` needed no edits.

## WASM build

`clap` is optional behind a new **default-on `cli` feature** (with `required-features = ["cli"]` on the binary), so the WASM build (`vcad-kernel-wasm` depends on vcad-render with `default-features = false`) does not pull clap in. Verified `cargo check -p vcad-render --no-default-features` still compiles.

## Verification

- `cargo test -p vcad-render` — passes
- Smoke-tested with a real fixture: stdout SVG, `-o out.jpg`/`.svg`/`-o -`, legacy `--jpeg`, bad-extension rejection, and a 3-file batch with one corrupt file (2 rendered, exit 2).

## Reviewer note

This is the smallest of a fan-out of 9 vcad-render improvement branches, and it restructures the CLI surface the others will add flags to (PNG output, camera, annotations, sections, sheet mode, etc.). **Landing this first** avoids each of those conflicting on the old hand-rolled parser. The extension→format map has a `TODO`-shaped seam where `.png` slots in once the PNG-output branch lands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7464,6 +7464,7 @@ dependencies = [
 name = "vcad-render"
 version = "0.9.3"
 dependencies = [
+ "clap",
  "image",
  "serde_json",
  "vcad-ecad-pcb",
--- a/changelog/entries/2026-07-15-vcad-render-cli-batch.json
+++ b/changelog/entries/2026-07-15-vcad-render-cli-batch.json
@@ -1,0 +1,9 @@
+{
+  "id": "2026-07-15-vcad-render-cli-batch",
+  "version": "0.9.4",
+  "date": "2026-07-15",
+  "category": "feat",
+  "title": "vcad-render CLI: -o output, batch mode, --help",
+  "summary": "The vcad-render CLI gains -o with extension-driven format, batch rendering of multiple files or a directory with --out-dir/--format, and proper --help.",
+  "features": ["render", "cli"]
+}
--- a/crates/vcad-render/Cargo.toml
+++ b/crates/vcad-render/Cargo.toml
@@ -6,9 +6,11 @@ publish = false
 description = "Project a .vcad document to static line art: isometric/orthographic SVG or raster JPEG."
 
 [features]
-default = ["raster"]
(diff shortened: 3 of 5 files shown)


ecto:
Thanks for the review. Addressed in `24b7086`:
- **Major (conflict check):** confirmed clap's `conflicts_with_all` rejects `-o file.svg --out-dir renders/` at parse time (exits with a usage error before `run()`), so the runtime guard is belt-and-suspenders for the multi-input case only.
- **Minor (batch_dest panic on `..`):** `batch_dest` is now fallible and surfaces a nameless path as a per-file batch error instead of `expect`-panicking.
- **Minor (progress on stderr):** kept on stderr by design so stdout stays clean for piped SVG; a `--quiet` flag is a reasonable future follow-up.

(The failing Rust CI was a `cargo fmt` miss, also fixed in that commit.)

PR #547, opened 2026-07-15
Animation timeline + verified video rendering for agents

ecto:
## What

vcad documents gain an optional **animation timeline** — a keyframed time axis over the parametric model itself — plus four MCP tools that turn it into the video rendering engine for agents:

| tool | role |
|---|---|
| `animate` | author/replace the timeline as data (params / joints / visibility / explode tracks + declarative camera shots), validated against the document |
| `timeline_from_simulation` | compile a recorded physics rollout (gym trajectory) into joint tracks — simulated motion becomes an animation source |
| `render_sequence` | timeline → **animated GLB** (real glTF animation channels); autoplays inline in the MCP viewer — the cheap dailies loop |
| `export_video` | kernel-rendered frames → GIF (gifenc, always) / MP4 (ffmpeg when present, artifact-offloaded), with the verification HUD burned into every frame |

## Why it's not just spectacle

When the document carries `clearance_specs`, both render tools **re-measure every spec across the sampled frames** and attach `{required_min_mm, observed_min_mm, worst_frame_t, holds}` — catching mid-motion violations endpoint checks miss (covered by a test). `export_video` burns the status into each frame: readouts left, green ✓ / red ✗ clearance right. The receipt ships inside the film.

## Pieces

- **IR**: `vcad-ir::animation` (`Timeline`/`AnimTrack`/`AnimKey`/`CameraShot`), optional `Document.timeline`, ts-rs regenerated
- **Engine**: `sampleSequence`/`poseDocument` sequencer, easing bit-matched to the Rust reference
- **GLB writer**: glTF animations (samplers/channels, STEP/LINEAR), meshless `extraNodes`, root wrapping; three.js round-trip tested
- **Camera**: shot intents compile to an invisible `__camera` carrier (yaw ⊗ pitch quaternion + dolly on scale); the viewer orbits its own camera from per-frame deltas — the model never rotates, and generic glTF players just ignore the node
- **Viewer**: autoplays glTF animation tracks (AnimationMixer), `#dev-glb` hostless harness
- **Fix found along the way**: `check_clearance` never worked on assembly-only documents (`partCandidates` only resolved scene roots) — instances now included with FK world transforms baked

## Verification

- 244 Rust (vcad-ir) + 111 engine + 839 mcp tests green; clippy/fmt clean; tool-surface fixture regenerated
- End-to-end demo committed at `examples/animation/gear-train.mjs` (3:1 gear pair, 360° turntable, 0.5 mm air gap verified across frames)
- Viewer playback verified in-browser: model motion, camera orbit/elevation/dolly, no console errors
- MP4 path verified live (ffmpeg encode → artifact URL)


ecto:
## Self-review sweep (8-angle multi-agent pass, adversarially verified)

Fixed in `dcd63147`:
- **HUD misattribution** (confirmed): with 2+ clearance specs the burned-in HUD paired spec[0]'s label with the min distance / verdict aggregated across all specs. Now shows the worst spec as a coherent unit (failing spec wins, then smallest margin).
- **Rollout duration off-by-one** (confirmed): `timeline_from_simulation` set `durationS = steps × stepTime` while the last key sits at `(steps−1) × stepTime`, freezing the final pose for one extra timestep.
- **Changelog summary** exceeded the schema's 200-char cap.

Known and deliberately deferred (design-level, worth discussing in review — none affect correctness):
1. **MP4 memory**: frames are buffered in RAM before piping to ffmpeg; a 600-frame 1280px export would be ~4GB resident. Fix is streaming frames to ffmpeg's stdin inside the render loop.
2. **Verification sweep cost**: `verifySequenceClearance` re-evaluates the whole scene per spec per frame (25×N evaluations); should pose+evaluate once per frame and reuse candidates across specs.
3. **poseDocument** structuredClones the whole document per frame (up to 600×/sweep); could clone only `parameters`/`joints`.
4. **Altitude**: `render_sequence` hand-attaches its `_meta` preview envelope instead of the dispatch layer owning animated previews (timestamp version defeats the viewer's version-keyed cache).
5. Thin-evenly sampling is hand-rolled 3× in animate.ts.

Also verified (refuted): the euler-z wrap at the 180° turntable crossing is harmless — the −2π yaw delta is an exact no-op through `sin`/`cos` in the spherical orbit.

PR #552, opened 2026-07-17
Tier 2: capacity-mesh global routing + route-to-tree termination (CM5 routability 0.929 → 0.933)

ecto:
Breaks the measured 0.93 plateau with evidence-driven architecture instead of more iteration. Full CM5 board, effort 1: **routability 0.921 (control) → 0.933 (best ever)**, stuck human-routed nets 45 → 34.

## The evidence chain

1. **Plateau measured**: five negotiation configurations (efforts 1–3, blanket/targeted rips, victim restore) all converged at ~0.93 — later rounds oscillated below the greedy+rip-up baseline, never above (visible in the new round logs).
2. **Ground truth interrogated** (`cm5_gap` example, new): for every net we couldn't route, diff how the *human* routed it. Verdict: the stuck nets received **1,956 vias — ~2/3 of the board's entire via budget** — with names like `VDD_BCM_CORE` (63 vias, 88 segments). They're power trees tapped wherever convenient, not pad-to-pad threads: our MST pad-pair decomposition was the binding constraint.
3. **SOTA surveyed** (arXiv/DAC/ICCAD 2023–26): multi-source/multi-sink tree routing is the settled fix (GAMER/Dr.CU/CUGR all route component-to-component); dedicated BGA escape (SAT/flow, MonoSAT-style) and TritonRoute-style escalating-window reroute are the ranked follow-ups.

## Changes

- **Route-to-tree termination**: the 3D search may terminate on any copper of the target pad's connected component (union-find over the net's live elements; overlap-exact goal cells; no endpoint via needed on tree copper). This is what moved the score.
- **Capacity-mesh global stage** (`router/global.rs`): GCell grid, free-space-derived capacity, soft-overflow A* with negotiation sweeps; corridors feed the detailed search as advisory windows (corridor-first, unbounded fallback). Plans 522 connections in 1ms. On this board it buys search locality, not routability — its telemetry (`0 overflowed cells` at 4mm) honestly says CM5 congestion is sub-cell, which is exactly the diagnostic the next arc needs.
- **`cm5_gap`** ground-truth analyzer: per-stuck-net layers/via-spans/length/region vs the human reference.
- `PcbLayer` gains `Ord`; session gains `for_each_of_net` + `copper_area_in`.

## What's left (documented for the next arc)

The remaining 34 nets are now dominated by signal pairs escaping the RP1 pin field (`ETH.*`, `USB3-1.DP`, `PCIe.TX_P`, `MDIO.*`) — the exact population for a dedicated SAT/flow escape phase, with escalating-window joint reroute as the cheaper second lever. Both are specced with paper references in the roadmap memory.

All 205 vcad-ecad-pcb tests green; clippy/fmt clean.


ecto:
@chojiai Addressed in 87c7ed9e + 4b33d75b:

- **CI failures**: root cause was `router/global.rs` never being staged (`git add -u` misses new files) — the module is now committed and all jobs should go green.
- **Minor (corridors in rip-up)**: implemented — `ripup_pass` now threads the global corridor map through to victim re-routes.
- **Major (union-find contact)**: documented rather than changed. `distance_to == 0` means the copper physically touches — metal-on-metal is electrically joined, while nearby-but-separate same-net islands keep `d > 0` and stay separate. The 0.01mm bbox slop is broadphase-only: it admits more pairs to the exact-contact check, it never unions them.
- **Major (tree-goal `< half_w`)**: documented rather than changed. The strict test is an overlap *certificate*: a goal cell guarantees the terminating trace physically overlaps tree copper. A sub-cell element marking zero cells degrades to the exact pad goal (safe); the suggested unconditional center-cell marking would let a route terminate up to pitch·√2/2 away from a small via — claiming connection to copper it never touches, which is the exact failure the certificate exists to prevent.

PR #569, opened 2026-07-17
particle: electron trapping is real — the neutralization dead-end reopened

ecto:
## What

Follows the merged e-injection "dead end" (naive electron injection recovered nothing) and **overturns it** with a direct measurement.

1. **`confinement` module** — traces an electron ensemble in the device's own cusp field and in the demagnetized copy (same electrostatics, coil currents zeroed), reporting the confinement-time enhancement, fate breakdown, and gyroradius.
2. **`orbitron_probe`** result at 100 kV, 45 mm rings, electrons launched at the mid-shell (near the rings, not the 0.2 inner shell the sweep used):

| shield A·t | enhancement | survivor | gyroradius |
|---|---|---|---|
| 0 | 1.0 | 0% | ∞ |
| 80 k | 121× | **75%** | 0.38 mm |
| 400 k+ | 140× (budget-capped) | **87.5%** | 0.03 mm |

**The two-ring cusp traps electrons hard** — 75–87% never leave (vs 100% expelled in 1.2 ns with B off), deeply magnetized (gyroradius 0.03 mm ≪ device). The e-injection sweep read ~0 because it launched electrons deep on-axis where the opposed coils cancel (B≈0) and E drives them straight out the axial point cusp before magnetizing. Launch location is everything — the exact polywell lesson, measured in our geometry.

3. **`neutralized` extended**: electron launch radius is now a parameter, with a generous electron flight budget and two decisive new diagnostics — `core_potential_change_v` (electron contribution to the *on-axis* ion well: negative = a virtual cathode that neutralizes the fusing core) and `electron_survivor_fraction`.
4. **`virtual_cathode`** example: sweeps electron launch radius and reports the core well-deepening + recovered ion yield — the decisive "does trapping actually help fusion" experiment (running; result folds into this PR).

Research log corrects the dead-end entry (append-only, the correction cites it).

## Verification

57 tests green; clippy `-D warnings` clean; fmt clean. Confinement tests verify B never worsens electron confinement, enhancement rises with ampere-turns and clears 1.2×, and gyroradius shrinks with field. Caveats carried on every result: non-relativistic Boris (the B-on/B-off ratio cancels most of it), single particles, enhancement budget-capped (a floor).

### DIFF
--- a/crates/vcad-kernel-particle/examples/neutralization_sweep.rs
+++ b/crates/vcad-kernel-particle/examples/neutralization_sweep.rs
@@ -52,6 +52,7 @@ fn main() {
             &topts,
             i_a,
             i_a,
+            0.5,
             &sc,
         )
         .expect("neutralized run");
--- a/crates/vcad-kernel-particle/examples/orbitron_probe.rs
+++ b/crates/vcad-kernel-particle/examples/orbitron_probe.rs
@@ -1,0 +1,66 @@
+//! Orbitron probe: does the shield's own cusp trap electrons?
+//!
+//! The e-injection dead end (electrons fall out of the −V well in one
+//! transit) has a caveat: the shield rings make a magnetic cusp, and near
+//! the wires electrons are magnetized. This sweep measures the electron
+//! confinement-time enhancement (full field vs the same launch with the
+//! coil currents zeroed) across shield ampere-turns — the load-bearing
+//! number for whether the neutralization lane has any architecture at all.
+//!
+//! Read with the caveats on [`vcad_kernel_particle::confinement`]:
+//! non-relativistic Boris (the B-on/B-off ratio cancels most of it),
+//! single particles, no electron self-fields, and the enhancement
+//! saturates at the flight budget (it is a floor, not the true trapping
+//! time).
+//!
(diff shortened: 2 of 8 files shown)


ecto:
**The arc's payoff — `virtual_cathode` + `virtual_cathode_current` results:**

Launch-radius sweep: electrons trap 71–94% at *every* radius, but at matched 30 mA the core well deepens only 0.2–0.3 kV — trapping ≠ neutralization; the lever is electron charge, not geometry.

Current sweep (30 mA ions, 100 kV): the virtual cathode **ignites** with electron current — core ΔV −0.2 → **−82.3 kV** and ion-yield gain 0.99 → **8.40×** (138× the amateur record) as e-current rises 30 mA → 10 A. The required e/i ratio (~100–300×) matches polywell practice.

Two asterisks carried on the result (and in the research log): e-e self-repulsion is unmodeled and dominant at high current (the perfect-injection bound is loose there — next module is electron-cloud self-consistency), and sustaining power = cusp-loss rate × injection energy (survivor fractions are budget-capped floors — loss accounting before any Q claim).

Arc: dead end (launch artifact) → trapping real (140×) → trapping ≠ neutralization → **neutralization real at polywell-scale currents**. Four self-corrections, all receipted in docs/research-log.md.

PR #635, opened 2026-07-23
fix(mcp): inline board preview after every PCB mutation

ecto:
## What

PCB-mutating tools (`route_nets`, `add_zone`, `set_placement`, etc.) now attach the layered board GLB to their result's `_meta` (`vcad.io/preview`), so the inline MCP widget renders the board after every copper mutation instead of showing "no geometry to preview".

## Why

The inline `_meta` preview from #626 only rode `mount: true` tool results. But `route_nets` / `add_zone` and the other copper mutators aren't mount tools, so their results carried no GLB — and Claude Code's widget `get_preview_glb` fallback isn't dependable. On a **board-only** session (no CAD part scene), that fallback is the only alternative, so the widget rendered "no geometry to preview" even though `describe_pcb` reported `preview_mesh_count: 16` and `render_pcb` worked.

The layered board GLB builder already existed (`pcbPreviewMeshes → buildGlb` in `preview.ts`, the same path the app's 3D board preview uses) — it just was never invoked for these tools.

## What changed

- **`server.ts`**: dispatch now attaches the inline board GLB for any geometry-writing tool when the session document contains a PCB (`sessionHasPcb`: a `PcbBoard` root or legacy `doc.pcb`), not just mount tools.
- **`ecad.ts`**: added the missing `behavior.geometry` flag to `add_zone`, `set_board_outline`, `set_placement`, `add_net_tie`/`delete_net_tie`, `add_via_array`, `route_diff_pair`, `length_match_traces` — so they carry a preview handle + version bump too.
- **`pcb-inline-preview.test.ts`**: extended the existing no-UI-capability regression test to call `route_nets` and `add_zone` after placement and assert each result carries a non-empty (`>1000` chars) `vcad.io/preview` GLB.
- Changelog entry (user-facing fix).

## Testing

Rebuilt the mcp dist first (workspace dist staleness), then ran the full mcp suite: **906 passed, 2 skipped, 0 failures**.

> [!NOTE]
> Reviewer gotcha surfaced while writing the test: `add_zone` layer names are `FCu`, not `F.Cu`.

### DIFF
--- a/changelog/entries/2026-07-23-pcb-mutation-inline-preview.json
+++ b/changelog/entries/2026-07-23-pcb-mutation-inline-preview.json
@@ -6,5 +6,5 @@
   "title": "Inline board preview after every PCB mutation",
   "summary": "route_nets, add_zone, and other copper mutations now attach the layered board GLB, so the inline widget no longer shows 'no geometry to preview' on board-only sessions.",
   "features": ["pcb", "mcp-apps", "preview"],
-  "mcpTools": ["route_nets", "add_zone", "set_placement", "set_board_outline"]
+  "mcpTools": ["route_nets", "add_zone", "set_placement", "set_board_outline", "add_net_tie", "delete_net_tie", "add_via_array", "route_diff_pair", "length_match_traces"]
 }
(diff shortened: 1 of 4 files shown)


ecto:
Addressed the minor finding in 992db3f — the changelog `mcpTools` array now includes `add_net_tie`, `delete_net_tie`, `add_via_array`, `route_diff_pair`, and `length_match_traces` alongside the four already listed. @chojiai

PR #636, opened 2026-07-23
fix(ecad): neck wide net classes into fine-pitch pads

ecto:
## Problem

The OHM autorouter (`route_nets`) silently failed any net whose class track width exceeds the target pad's narrow dimension. A 1.0mm power class landing on a 0.35mm TSSOP-16 pad at 0.65mm pitch can never probe clearance-legal at the endpoint, so every maze search died at the pad and the net came back as a bare `UnconnectedNet` with no explanation.

**Repro:** RP2040 motor board — DRV8833 on a TSSOP-16 (0.65mm pitch, ~0.35mm pads) with the power class at 1.0mm. VM and every motor-output net returned `UnconnectedNet`; dropping the class to 0.5mm only partially fixed it.

## Fix

Two parts, both in `crates/vcad-ecad-pcb/src/router/auto.rs`:

**1. Pad-entry neck-down.** `search_route` now plans a per-endpoint neckdown via a new `plan_neckdown`: when the class width can't land on a pad, it fans a ~1mm approach stub out of the pad (trying outward-from-footprint directions first — out of the pin row on a QFP/TSSOP) at `min(pad width, class width)`, requiring both the stub and a full-width landing disc to probe clearance-legal, then runs the maze at full class width from the landing point. The stub rides through the existing `Candidate::thin_segments` neck-down channel (already used by pair connectors), so it's re-probed at commit time — **the DRC-clean invariant is untouched**. The push-shove fallback uses the landing points too.

**2. Named diagnostics.** `diagnose_unrouted` checks the pad-landing cause before the congestion census. When an endpoint's class width can't land and no legal neck approach exists, the reason now reads e.g. `class width 1.00mm cannot land on U2.12 pad 0.35mm and no clearance-legal neck-down approach exists — reduce the class width or free space around the pad`, instead of generic congestion output. This surfaces through `route_nets`' existing `diagnostics` field.

## Tests

- `wide_class_necks_down_into_fine_pitch_pad` — the repro: 1.0mm class into a 0.35mm pad flanked by other-net pads at 0.65mm pitch. Asserts the net routes, the run stays at 1.0mm, the pad-touching trace is ≤0.35mm and ≤1mm long, and the applied board is short/clearance clean.
- `unroutable_wide_class_names_pad_landing_cause` — pad sealed by a ring of other-net pads; asserts the net fails **with** a diagnostic naming the pad, the pad width, and the class width.

All 236 crate tests pass; `cargo clippy -p vcad-ecad-pcb -- -D warnings` and `cargo fmt --check` are clean.

## Reviewer notes

The escape/fan-out rescue stubs still commit at full class width — extending neckdown to dog-bone fanouts is a possible follow-up, not covered here.

### DIFF
--- a/changelog/entries/2026-07-23-router-pad-entry-neckdown.json
+++ b/changelog/entries/2026-07-23-router-pad-entry-neckdown.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-23-router-pad-entry-neckdown",
+  "version": "0.9.4",
+  "date": "2026-07-23",
+  "category": "fix",
+  "title": "Autorouter necks wide net classes into fine-pitch pads",
+  "summary": "A net class wider than its target pad now necks down for the final pad approach instead of being silently unroutable, and a net that still fails names the pad-landing cause in its diagnostic.",
+  "features": ["ecad", "autorouter"],
+  "mcpTools": ["route_nets"]
+}
--- a/crates/vcad-ecad-pcb/src/router/auto.rs
+++ b/crates/vcad-ecad-pcb/src/router/auto.rs
@@ -14,7 +14,7 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 
 use rayon::prelude::*;
 
-use vcad_ir::ecad::{Pcb, PcbLayer};
+use vcad_ir::ecad::{Footprint, Pad, PadShape, Pcb, PcbLayer};
 use vcad_ir::Vec2;
 
 use crate::ratsnest::{compute_ratsnest, NetConnection, Netlist, NetlistNet, RatsnestLine};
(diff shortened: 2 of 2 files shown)


ecto:
Thanks @chojiai — reviewed all three findings against the committed code; none require a change:

- **Stub gap (Major):** Not reachable. `validate_and_commit` probes the maze `segments`, the neck `thin_segments`, and the vias, and returns `None` mutating nothing if *any* probe fails — the stub and the maze trace are committed atomically in the same call, so there's no interleaving where the trace lands but the pad-to-landing stub is dropped.
- **Double `plan_neckdown` (Major):** Intentional. `diagnose_unrouted` runs against the final settled session because that's the state that explains why the net can't route *now*; consuming a stale search-time verdict would be less accurate, not more.
- **Epsilon (Minor):** The suggested larger/relative epsilon makes neck-down trigger *less* often, which reintroduces the exact silent-unroutable bug this PR fixes. `pad_min_dim` reads pad-shape dimensions directly from the IR (no coordinate-transform accumulation), so the tight inclusive guard `w > pad_w + 1e-9` is the safe direction.

PR #637, opened 2026-07-23
fix(pcb): maze-rescue plane stitching for boxed-in fine-pitch pads

ecto:
## What

When a pad on a copper-plane net can't take an at-pad stitching via and the radial dog-bone escape finds no *straight* stub out, `route_nets` now maze-routes the pad to the nearest same-net stitched point (a stitching via, or a pad already sitting on its plane) using the full rescue arsenal — instead of emitting an `UnstitchedPad` DRC violation. `UnstitchedPad` is now reserved for the truly impossible cases the maze also can't reach.

## Why

Repro: the RP2040 motor board with a bottom GND plane left the DRV8833's GND and AISEN/BISEN pads (fine-pitch TSSOP) floating, each surfacing as an `UnstitchedPad` error whose message already computes an escape vector.

Root cause was **not** a missing escape mechanism. `escape_endpoint`'s dog-bone fanout already exists, but it only probes *straight* stubs radiating from the pad. A fine-pitch pad fenced in by already-routed neighbour fanouts fails every one of the 160 ring candidates (in the new test, 159 rejected on the stub-legality probe). A *bent* path threads out fine; the maze router can find it, the radial search can't.

## Changes

- **`crates/vcad-ecad-pcb/src/router/auto.rs`** — after `stitch_planes`, failed pads get a maze rescue: `try_route` from the pad to the 3 nearest same-net stitched points, with bends / pin-field escape / shove / self-dropped vias. Any path found reaches the plane galvanically. Only pads that still fail are pushed back to `failed_pads` → `UnstitchedPad`.
- Made the fine-pitch fanout threshold inclusive (`pitch <= FINE_PITCH_MM`) so exactly-0.65 mm TSSOPs are forced off-pad, matching the constant's own documentation.
- Regression test `stitches_tssop_plane_pads_via_escape_fanout`: a 0.65 mm-pitch TSSOP with GND + adjacent same-net sense pads, mixed-net neighbours, a bottom GND plane, and a contested fanout region. Asserts zero `UnstitchedPad`, zero short/clearance violations after routing, and that no stitch via lands on a fine-pitch pad. Fails without the fix, passes with it.

## Testing

- `cargo test -p vcad-ecad-pcb --lib` — 235 passed, 0 failed
- `cargo clippy -p vcad-ecad-pcb -- -D warnings` — clean (CI's lib-only clippy)
- `cargo fmt -p vcad-ecad-pcb --check` — clean

Note: `cargo clippy --all-targets` has two pre-existing failures in files this PR doesn't touch (`pair.rs` unused import, `si_claims.rs` items-after-test-module); CI runs lib-only clippy, which is clean.

### DIFF
--- a/changelog/entries/2026-07-23-plane-stitch-maze-rescue.json
+++ b/changelog/entries/2026-07-23-plane-stitch-maze-rescue.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-23-plane-stitch-maze-rescue",
+  "version": "0.9.4",
+  "date": "2026-07-23",
+  "category": "fix",
+  "title": "Plane stitching escapes boxed-in fine-pitch pads",
+  "summary": "route_nets now maze-routes a bent escape to the plane when a fine-pitch pad can't take an at-pad or straight dog-bone stitching via, instead of reporting UnstitchedPad.",
+  "features": ["pcb", "autorouter", "drc"],
+  "mcpTools": ["route_nets", "run_drc"]
+}
--- a/crates/vcad-ecad-pcb/src/router/auto.rs
+++ b/crates/vcad-ecad-pcb/src/router/auto.rs
@@ -458,10 +458,11 @@ pub fn route_all_with_opts(
     // Maze rescue for pads the radial dog-bone couldn't escape (issue: a
     // fine-pitch pad boxed in by already-routed copper has no *straight*
     // stub out, but a bent path threads it). Route the pad to the nearest
-    // same-net stitched point — a stitching via or a pad sitting directly on
-    // its plane — with the full maze arsenal; any path it finds (including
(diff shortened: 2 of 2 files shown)


ecto:
@chojiai addressed the review findings in 6017e362:

- **Same-net traces on plane layer as rescue targets** — now included: the rescue collects endpoints of same-net segments already routed onto a plane layer, alongside stitch vias and on-plane pads.
- **Pad-to-itself routing** — filtered with `targets.retain(|t| dist(*t, pad_pt) > 1e-6)` before the sort, so a coincident same-net target or zero-length route is never attempted.
- **`escape_endpoint` doc comment** — verified the full doc block already reads "0.65 mm-pitch-and-finer pads"; no lingering "sub-0.65 mm" wording remained.

235 lib tests + lib clippy + fmt clean.

PR #672, opened 2026-07-25
fix(router): enforce via hole-to-hole spacing at probe time

ecto:
## The hole

`RouteSession::probe` filters candidates by layer (`e.layer != layer`) because copper is indexed per layer. Two vias whose layer spans **do not overlap** share no copper element, so the oracle never compared them — while their drill barrels, which are layer-independent, could physically collide. Every clearance check passed; the board was unmanufacturable.

Evidence from the 2026-07-24 CM5 fab campaign: a routed board carried 9 cross-net via pairs closer than the hole-to-hole rule, worst `-0.024mm` overlap (`/MIPI0.D3_N` vs `/USB3-0.DM`). They were caught only by a post-hoc python scan and fixed by strip-and-reroute. `router::legalize` does have a fail-closed `HoleToHole` demote path, but it fixes one violation per round under `MAX_FIX_ROUNDS = 4` — nine of them outran it.

## The fix

A second index in the session, over **drilled holes** rather than copper:

- `probe_hole` / `probe_via_hole` — mirrors `drc::check_hole_to_hole` exactly: illegal when `dist(centers) - rA - rB < rules.hole_to_hole - 1e-6`, **regardless of layer span**. Seeded from `pcb.vias` (any span) and through-hole pad drills.
- `commit_hole` — hole ids share the `SpanId` space, so `remove` rips a via's barrel out with its copper and rip-up-and-reroute keeps working. Separate tombstone counter and compaction per index.
- `via_drill_for` — per-net-class drill diameter, mirroring `legalize::via_geom_for`, i.e. how the router sizes a via on write-back. No drill plumbing needed at call sites.

Wired into every via-placement path:

| site | what it gates |
|---|---|
| `maze::via_ok` | in-search, so A* *avoids* bad via sites instead of failing at commit |
| `auto::escape_endpoint::via_legal` | escape / dog-bone / plane-stitch vias |
| `auto::validate_and_commit` | the commit gate — last line of defense |
| `complete::edge_ok` | the complete window router's via edges |
| `auto::commit_via` | now also commits the barrel, so the index stays complete |

Two deliberate semantic choices, both documented in the code:

- **Same-net holes are exempt.** Coincident same-net drills are the *merge* case, already owned by `legalize::merge_same_net_vias` — refusing them at probe time would break at-pad stitching.
- **Net ties are NOT exempt.** A tie is an electrical exemption; a drill is mechanical. This matches the DRC pass exactly, which is the point.

Pad drills are included, not just vias. The task's evidence was via-vs-via, but it's the same rule and the DRC already flags via-vs-pad — excluding pads would leave router and DRC disagreeing, which is the failure mode being fixed.

## Verification

- **New router test** `via_refused_when_drills_collide_across_disjoint_layer_spans`: a blind `FCu..In1Cu` via and a buried `In2Cu..BCu` via 0.5mm apart on a 4-layer board. Asserts every copper probe passes (that *is* the hole), the hole probe reports the 0.1mm gap and names the blocker, `validate_and_commit` refuses and `session.len()` is unchanged, a 1.0mm placement is accepted, and `check_drc` flags the same pair as `HoleToHole` — so the session and the DRC agree.
- Two session tests: rule arithmetic + boundary + same-net exemption + rip-up reopening the spot, and via-vs-through-hole-pad.
- `cargo test -p vcad-ecad-pcb --features gpu` — 255 passed, 0 failed.
- `cargo clippy -p vcad-ecad-pcb --features gpu -- -D warnings` — clean. (`--all-targets` has 2 pre-existing `items-after-test-module` errors, present on `main` at `1e69b5a`.)
- **40-net CM5 subset**, built `--release --features gpu`: output is identical to the pre-fix baseline — 1292 segments, 84 vias, 39 routed / 1 unrouted (the same pre-existing plane-stitch failure on `Net-(C240-Pad2)`) — with zero cross-net via pairs under the rule either way. No routability or wall-clock regression; the hole query is one small R-tree lookup per candidate via.
- **Full CM5 route is running now** (~2h). I'll post the cross-net via-pair scan of `/tmp/out.pcb.json` on this PR when it lands.

## Reviewer note: the diff-pair test board changed

`router::pair`'s `pair_board` declared `hole_to_hole: 0.5` with `via_drill: 0.2` on legs `gap + width = 0.45mm` apart. Two 0.2mm drills 0.45mm apart leave a 0.25mm gap — under that rule **no** pair layer transition on that board is manufacturable, and the two transition repros only passed because the oracle couldn't see it. The board now declares the 0.2mm rule its own geometry implies, and both repros assert `assert_twin_holes_clear` so the invariant is checked rather than assumed. CM5 itself uses `holeToHole: 0.25`, so real pair transitions are unaffected.


ecto:
## Acceptance criterion 2: full CM5 route — **zero cross-net via pairs**

Fresh full-board route on this branch, no post-pass:

```
cargo build --release -p vcad-ecad-pcb --features gpu --example cm5_bench
RUST_LOG=warn VCAD_GPU_BATCH=1 VCAD_GPU_NEGOTIATE=1 \
  target/release/examples/cm5_bench .scratch/CM5RevEng.kicad_pcb 1 999999 /tmp/holefix_out.pcb.json
```

```
board: 10 copper layers, 436 nets, 479 footprints, 3037 pads
human: 8115 segments, 2902 vias, 7667 mm copper, 409 routed nets
vcad:  7328 segments, 741 vias, 6052 mm copper, 396 routed / 12 unrouted nets
score: routability 0.987, via ratio 0.26x human, length ratio 0.79x human, 6807.4s
```

Scan over the `vias` array (position.x/y, drill, net), same arithmetic as `drc::check_hole_to_hole`:

```
/tmp/holefix_out.pcb.json: 741 vias, holeToHole = 0.25mm
cross-net via pairs under the rule: 0
```

**0 violations across 741 vias.** The 2026-07-24 fab campaign found 9 such pairs on a comparable route — worst `-0.024mm` (`/MIPI0.D3_N` vs `/USB3-0.DM`) — caught only by a post-hoc python scan and fixed by strip-and-reroute. They are now refused at probe time.

Routability 0.987 is at or above the 0.985 recorded best (#669), so the extra constraint did not cost coverage. The 40-net iteration subset also routes byte-identically to the pre-fix baseline (1292 segments, 84 vias, same single pre-existing plane-stitch failure), which is the wall-clock evidence: the hole query is one small R-tree lookup per candidate via.

Also merged current main in — the three unrelated phyz breaks (#673, #675) that were blocking every job before this branch's code could even compile.

PR #674, opened 2026-07-25
feat(native): vcad for Apple Vision Pro

ecto:
Ports the full native editor to visionOS, plus two macOS fixes found along the way.

## What changed

**One codebase, two targets.** The mac app's sources (`Document`, `Editor`, `Kernel`, `Materials`, `IntentBar`/`IntentClient`, `ReceiptLedger`, `Chime`, `UsdzExport`, `Shell` panels) are **symlinked** into `apple/VcadVision/Sources/VcadVision/Shared/` and compiled for both platforms — no fork, no copy to drift. Mac-only stretches sit behind `#if os(macOS)`:

- `EditorView` + `DocumentMenu` (NSSavePanel/NSOpenPanel)
- `ViewportView` and its extensions (camera-driven `RealityViewCameraContent`, NSEvent gestures, picking, hover)
- release-to-desktop window plumbing
- NSEvent key/scroll monitors, `NSHapticFeedbackManager`, the raytrace overlay

`PlatformShims.swift` maps `NSColor` → `UIColor`; `Color(portedColor:)` and `.onEscape` bridge the remaining API spelling gaps.

**`DocumentVolume` — the visionOS viewport.** Full document scene in a volumetric window: parts, assembly instances, shared-mesh pattern instancing (same `part<i>_inst<j>` naming as the studio so shared helpers keep working), per-part materials, the CAD feature-edge overlay, orange selection emissive, and the transform gizmo. Interaction is spatial:

- look-and-tap to select
- **long-press** toggles the boolean multi-selection (a volume has no modifier keys, so this is the ⌘-click)
- pinch-drag spins the part
- parameter scrubs re-solve and swap meshes **in place** rather than rebuilding the tree
- documents open via `.fileImporter` — the port of the mac `NSOpenPanel` path

Feature tree, inspector/receipt, tool palette, and the AI composer ride ornaments around the volume.

**macOS fixes** (same session, same files):
- Release-to-desktop drops the desktop-snapshot skymap and its ScreenCaptureKit permission prompt; it now uses the studio IBL plus an explicit key/rim/fill rig. Without that rig the floating parts rendered as black silhouettes — caught on screen and fixed.
- The Object Inspector floats **beside the part it inspects**, projecting the selected part's center to overlay coords and tracking it through orbits (parks under the palette when nothing is selected).

## Build plumbing (the interesting yaks)

- `aarch64-apple-visionos-sim` is a **tier-3** Rust target → `build-ffi.sh` uses nightly + `-Zbuild-std=std,panic_abort` against the `xrsimulator` SDK.
- `clipper-sys` 0.7.2's `build.rs` **panics** on visionOS (`not implemented: target_os: visionos`). Vendored under `third_party/clipper-sys` with the missing libc++ link arms added, wired via `[patch.crates-io]`. Source is otherwise identical to the registry crate.
- SwiftPM can't express a visionOS app target, and `swift build --triple arm64-apple-xros…` emits ELF-style linker flags (`-rpath=$ORIGIN`). `bundle.sh` builds via `xcodebuild` and assembles the `.app` by hand, then `simctl install`/`launch`.

## Verification

- `pattern-test.vcad` renders in the visionOS 26.5 simulator with the feature tree, inspector (material picker + live measurements), palette, composer, gizmo, and selection all live
- macOS app builds and launches unchanged
- `cargo check -p vcad-ffi` green on the host target (the `[patch.crates-io]` doesn't disturb normal builds)

## Reviewer notes

- The symlinked `Shared/` directory is the load-bearing idea here — worth a look at whether you'd rather see a shared SwiftPM target long-term. Symlinks kept the diff honest (no duplicated source) without restructuring the mac package.
- `third_party/clipper-sys` is a vendored fork of a published crate. It carries its own `[workspace]` so cargo doesn't try to adopt it.
- Build artifacts (`.build/`, `dist/`, `Libs/`) are gitignored for the new target, matching the existing `VcadApp` rules.

**Known gaps** (deliberate, not blockers): sketch-mode tap-to-place isn't wired into the volume yet, and the gizmo renders but isn't draggable there — both need a spatial-ray equivalent of the mac's `kernelRay`.


ecto:
Heads-up on scope creep, all of it forced by upstream `phyz` drift rather than by this branch:

1. **`..Default::default()` on the M11 `Joint` literals** — collided with #673, which landed the identical fix while this was in CI. Resolved by taking main's version verbatim; those two files are now byte-identical to main.
2. **`phyz-{gpu,model,math,md}` bumped `0.1` → `0.3`** (latest commit). phyz moved every crate onto its shared workspace version (0.3.1), so the sub-crates are no longer independently `0.1.x` and the requirement stopped resolving:
   ```
   error: failed to select a version for the requirement `phyz-md = "^0.1"`
   ```
   This is what the comment above those deps asks for ("bump these when the sibling repo cuts a release"). **main's tip `049b40ad` is currently red for exactly this** — it is not specific to this PR.

⚠️ **This requires a local `phyz` checkout at 0.3.x.** A sibling repo still on the old workspace version (0.1.0) will now fail to resolve locally — same staleness CI has already moved past.

The deeper issue: `ci.yml` does `git clone --depth 1 https://github.com/ecto/phyz.git ../phyz`, unpinned. Every phyz shape change breaks every vcad PR simultaneously, and today it did so twice in an hour. Pinning that clone to a SHA (bumped deliberately) would make CI hermetic. Out of scope here, but worth its own PR.

ecto:
**Blocked on upstream, not on this branch.** Parking CI here.

`Rust (stable)`, `Rust (nightly)` and `WASM build` all fail the same way — inside **phyz**, not vcad:

```
error[E0277]: the trait bound `ExprId: Scalar` is not satisfied   (×72)
error: could not compile `phyz-diff` (lib) due to 72 previous errors
```

`ExprId` is tang's, `Scalar` is phyz's. `ci.yml` clones `tang`, `phyz` and `loon` each at unpinned `--depth 1` HEAD, and phyz HEAD no longer compiles against tang HEAD. Nothing inside vcad can fix that — it needs a phyz-side (or tang-side) fix, or pinned SHAs in `ci.yml`.

Progress from the two fixes already on this branch: **KiCad export verification** and **Torture track** went red → green once phyz resolved again. What's left is only the tang↔phyz incompatibility.

This PR's own content is verified outside CI:
- macOS app builds and launches
- visionOS `BUILD SUCCEEDED`, runs in the visionOS 26.5 simulator with `pattern-test.vcad` rendering, selection and panels live
- `cargo check -p vcad-kernel-physics --tests` passes

Ready to re-run CI once phyz is fixed upstream.

PR #677, opened 2026-07-25
Close the CM5 autorouting verdict's last unknowns (0 unknown, per-connection certificates)

ecto:
Every connection on the best CM5 board now ends **ROUTED** or **PROVED-INFEASIBLE** — no unknowns — and every certificate names the copper forming its cut.

## Result

| run | verdict |
|---|---|
| before (same input, same driver contract) | routed 43 / proved 21 / **unknown 31** |
| after | routed 44 / proved 51 / **unknown 0** |
| second pass over the output | routed 0 / proved 45 / unknown 0 (fixed point) |

The 31 unknowns were 28 cluster-level budget trips, 1 singleton budget trip, and 2 paths the fail-closed oracle rejected — budget and discretization artifacts, as expected.

## What closes them

- **k = 1 is decided by reachability, not the exhaustive DFS.** A lone connection cannot trip a budget: BFS returns a shortest path, or an exhausted reachable component that *is* the proof. This is what makes splitting a cluster terminate.
- **A sequential-BFS witness pass runs before the DFS** for joint instances. The DFS's first descent is a greedy walk, not a shortest path, so on a wide window it could wander for millions of expansions before its first completion — the dominant source of unknowns. Success is a witness (node-disjoint by construction); failure proves nothing and falls through, so completeness is untouched.
- **`WindowBudget` couples the expansion budget to a cells-per-axis cap**, so a wider window keeps its pitch at the `width + separation` floor instead of coarsening until unrelated terminals collide in one cell. (Lever 2, done the right way: cap raised *with* the cell budget.)
- **The driver escalates instead of accepting an unknown**: joint at 1x/5x/25x budget, then per connection in its own 2 / 8 / 20 mm window. A joint infeasibility is no longer charged to the individual connections — it says only that the k cannot all be routed *at once*, so each connection earns its own verdict.
- **Pair-aware pitch**: twins in one window are spaced by their declared gap, restoring "distinct cells ⇒ mutually legal" for pairs (lever 3).

Certificates got sharper, not weaker. The severed case reports whichever terminal's pocket is smaller, with the enclosing cut's node count, extent, and the nets holding it; the walled-in case names the pad's own layers and its blockers. All 51 name real copper:

> net `/LPDDR4 RAM/CK_C_B` is severed inside the window: breadth-first search from its from terminal at (128.65, 77.15) exhausted every reachable node — 5 of the 50108 clearance-free (cell, layer) nodes on the 10-layer stack — without touching the other terminal at (133.55, 66.42) (layer FCu); the enclosing cut is 14 blocked nodes spanning x=128.22..129.41, y=76.08..77.57, held by `/LPDDR4 RAM/CA5_B` (10), `/LPDDR4 RAM/CK_T_B` (10), `/LPDDR4 RAM/CA2_B` (8). With k = 1 reachability is exact, so no path exists on the canonical grid at pitch 0.297 mm

## Three legality gaps closed in the session oracle

Routing more connections surfaced three ways "routed" copper could still fail the board. Fixing them is what keeps the new copper DRC-clean:

- **`probe_drill` / `commit_drill`** — hole-to-hole across layer spans and nets. Two vias on disjoint spans share no copper layer, so no layer-scoped probe ever compared them; they landed as drill collisions (the durable fix for a known open item). A via barrel is also now probed on *every* layer it spans, and consecutive transitions at one point merge into one barrel instead of stacking drills at zero spacing.
- **`TerminalLayers`** — terminals can be pinned to the layers their net actually has copper on, with a probed dog-bone escape (stub on the pad's layer, barrel checked across its span) when the search wants a different layer. Previously a stub took the first free layer, which on a 10-layer board is usually not the pad's: 142 of 162 new pad-touching stub ends were electrically dangling.
- **`set_strict_pair_coupling`** — scores the intra-pair gap the way the DRC does, pads and via annuli included, with the DRC's 1.5 mm pad-breakout exemption mirrored. **Opt-in**, because it is stricter than the geometry the pair router realizes today (its dog-bone offsets and jogs are sized to base clearance); the verdict driver turns it on and pays for it in routability. Flipping the default needs the pair router's geometry fixed first — noted below.

## DRC (same input board, `drc_json`)

| rule | base | after |
|---|---|---|
| Clearance | 74 | **74** (baseline floor, exactly) |
| Short | 1315 | **1315** |
| MinTraceWidth | 69 | **69** |
| HoleToHole | 0 | **0** |
| UnconnectedNet | 252 | **223** |
| NetIslands | 18 | 30 |
| total | 1813 | **1796** |

No new route-attributable violations. NetIslands +12 is an accounting shift: all 19 newly-listed nets had *zero* copper on the base board (they were UnconnectedNet entries) and now carry partial copper — some MST edges routed, the rest proved infeasible. An independent connectivity check over all 41 touched nets finds **zero floating groups**: every committed path is galvanically attached to its net's pads or pre-existing copper, with no zero-length traces and no coincident vias.

## Verification

- `cargo test -p vcad-ecad-pcb` — 252 passed, 0 failed (4 new tests: lone connection never unknown, severed connection proved with a named cut, pinned terminals attach on their own layer, drill probe sees holes no copper probe could).
- `cargo clippy -p vcad-ecad-pcb --features gpu -- -D warnings` — clean. (`--all-targets` reports two `items after a test module` lints in `router/auto.rs` and `router/si_claims.rs`; both are pre-existing on main, confirmed by stashing.)

## Follow-up (not in this PR)

The pair router emits dog-bone offsets and jogs that come a gap-width short of the DRC's pair rule (a leg trace 0.175 mm from its twin's via, where the DRC requires 0.245 mm mid-board). Repro: `router::pair::tests::pair_layer_transition_keeps_twin_clearance` with `set_strict_pair_coupling(true)`. Fixing that geometry is what would let strict coupling become the default.

### DIFF
--- a/changelog/entries/2026-07-24-router-drill-aware-vias-pad-layer-stubs.json
+++ b/changelog/entries/2026-07-24-router-drill-aware-vias-pad-layer-stubs.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-24-router-drill-aware-vias-pad-layer-stubs",
+  "version": "0.9.4",
+  "date": "2026-07-24",
+  "category": "fix",
+  "title": "Router vias clear drills; stubs leave pads on the pad's layer",
+  "summary": "The routing oracle now checks hole-to-hole spacing across layer spans (two vias on disjoint spans no longer collide in the drill file), probes a via barrel on every layer it spans, and can pin a connection's terminals so stubs leave a pad on a layer the pad is actually on instead of floating on an inner layer.",
+  "features": ["pcb", "autorouter", "drc"],
+  "mcpTools": ["route_nets", "run_drc", "validate_for_fab"]
+}
--- a/crates/vcad-ecad-pcb/examples/cm5_verdict.rs
+++ b/crates/vcad-ecad-pcb/examples/cm5_verdict.rs
@@ -1,22 +1,80 @@
 //! Verdict driver: put the still-unrouted connections of a routed board in
 //! front of the COMPLETE window router, cluster by cluster, and demand an
-//! answer — Routed (commit-quality paths), ProvedInfeasible (bottleneck-cut
-//! certificate), or BudgetExhausted (honest unknown). The campaign's closing
-//! argument: every connection ends accounted for.
+//! answer — Routed (commit-quality paths) or ProvedInfeasible (bottleneck-cut
+//! certificate). The campaign's closing argument: every connection ends
+//! accounted for, and nothing ends "unknown".
+//!
+//! The ladder, per cluster, escalates rather than accepting an unknown:
+//!
+//! 1. **joint window, base budget** — the k connections decided together;
+//! 2. **joint window, 5× then 25× budget**, cells-per-axis raised with the last
+//!    rung so a wider search keeps its pitch at the `width + separation` floor
+//!    instead of coarsening until unrelated terminals collide in one cell;
+//! 3. **per-connection endgame** — whatever is left is decided one connection at
+//!    a time in its own window: 2 mm, then 8 mm, then 20 mm of margin, each with
+//!    a cell budget that holds the pitch at that floor. A lone connection is
+//!    settled by reachability, which is exact and cannot trip a budget: it
+//!    routes, or its severed reachable component *is* the proof.
+//!
+//! Splitting is not a fallback for weak proofs, it is what makes the
+//! certificates *per connection*: a joint infeasibility says only that the k
+//! cannot all be routed at once, so it is never charged to the individual
+//! connections — they each get their own verdict. And because each rung searches
+//! a strictly larger space than the last, the certificate that survives to the
+//! end is the strongest one available.
+//!
+//! Everything commits fail-closed through the session oracle — traces, via
+//! barrels on every layer they span, drills against the hole-to-hole rule, and
+//! diff-pair coupling scored the way the board's DRC scores it.
 //!
 //! ```bash
 //! cargo run --release -p vcad-ecad-pcb --example cm5_verdict -- routed.pcb.json [budget] [out.pcb.json] [max_cluster]
 //! ```
+//!
+//! `budget` is DFS expansions; a value below 1000 is read as a multiplier of
+//! 1e6 expansions (so `3` and `3000000` mean the same thing).
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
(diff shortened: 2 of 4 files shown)


ecto:
**Rebased onto `main` (`a5a57e4a`)** — the four red Rust/WASM jobs were not this branch. They died at dependency *resolution* (`failed to select a version for the requirement phyz-md = "^0.1"`, candidate 0.3.1) because CI clones the sibling `phyz` repo at its moving main tip, which had bumped to 0.3.1. #675 fixed that on `main` 20 minutes after this PR opened; the rebase picks it up. WASM build, KiCad export verification, and both audits are green again on the rebased head.

Verified locally against `phyz` at its actual `origin/main` (99778e6, the state CI sees) using a detached phyz worktree with the path deps temporarily repointed, since this machine's phyz checkout lags main:

- `cargo clippy --workspace --exclude vcad-desktop --features vcad-kernel-text/no-builtin-font -- -D warnings` — clean (the exact command the failing job runs)
- `cargo check --workspace --exclude vcad-desktop --all-targets` — clean
- `cargo test -p vcad-ecad-pcb` — 252 passed
- board result unchanged after the rebase: routed 44 / proved-infeasible 51 / unknown 0, `Clearance 74` / `Short 1315` / `HoleToHole 0` all at the baseline floor

### Overlap worth resolving before merge

**#672 (`fix(router): enforce via hole-to-hole spacing at probe time`)** independently adds the same capability to `RouteSession`, with a broader API (`probe_hole` / `probe_via_hole` / `commit_hole`, per-net via drill from net classes, blocker reporting, wired into `auto.rs` / `maze.rs` / `pair.rs` / `complete.rs`). This PR's narrower `probe_drill` / `commit_drill` will conflict with it in `session.rs`. #672's is the better base; the two things worth porting from here are the grid-pitch floor at `drill + hole_to_hole` in `router/complete.rs` and the `ViaClass`-driven barrel probe on every layer a via spans.

One substantive difference to settle either way: #672 skips same-net holes (`h.net == net`), while `drc::check_hole_to_hole` is net-agnostic — it compares every via barrel and through-hole pad drill with no net filter, so two same-net vias closer than `rules.hole_to_hole` but not coincident are a real violation a same-net-exempt probe won't catch. `probe_drill` here is net-agnostic to match the DRC.

**#681 (`universal pair coupling and the SI receipt`)** looks like it supersedes the opt-in `set_strict_pair_coupling` added here, and may well fix the pair-router dog-bone/jog geometry that forced it to be opt-in (see the follow-up note in the PR description). If #681 lands first, this flag should probably become the default or disappear.

PR #690, opened 2026-07-25
docs(router): correct every CM5 DRC/SI claim measured against the inflated baseline

ecto:
Re-measures and corrects every published DRC and signal-integrity claim about the CM5 autorouting benchmark. All of them were computed against a stripped-fixture baseline that one of our own bugs had inflated.

## Why

KiCad stores a pad's angle as **absolute** — it already includes the footprint's orientation. vcad's importer stored that raw value while eleven consumers compose `fp.rotation + pad.rotation`, double-counting the footprint rotation on every rotated part. On fine-pitch rotated packages the neighbouring pads **overlapped**, manufacturing phantom shorts and clearance violations out of nothing (#684).

Those phantoms landed in the **baseline**, which is the denominator of every route-attributable claim we publish. A bigger floor makes our delta look smaller, so the error flattered us in every published comparison — the direction that most needs correcting.

## Method

Re-measured from scratch on merged main with #684 rebased in. DRC and SI are deterministic given a fixed board file, so each was run **twice**; every pair agreed exactly (no spread to report). The route stage is the only stochastic step and is reported with its wall-clock.

Note: the full-board route shared the machine with another session's full-board route (load average 23.4). Contention only inflates wall-clock, so the `<30 min` conclusion holds a fortiori.

### The baseline itself (stripped fixture, all copper removed)

| | before | after |
|---|---|---|
| short/clearance | 980 | **311** |
| `Clearance` | 74 | **23** |
| `Short` | ~906 | **258** |

**648 violations this project attributed to the reverse-engineered source were ours.**

### The human production board

| | before | after |
|---|---|---|
| total violations | 16,485 | **14,104** |
| short/clearance | — | **6,994** |

### Our board — the claim that broke

| | before | after |
|---|---|---|
| `Clearance` | 74 (claimed = the floor) | **173** |
| floor | 74 | **23** |
| **route-attributable `Clearance`** | **0** | **150** |

**"Route-attributable violations: ZERO" was false.** It read `74 − 74 = 0` because the inflated floor happened to equal what our board scored.

The shipped fab package re-measures at `Clearance` **81** vs floor 23 = **58 route-attributable**.

### SI receipts

| 40-net subset | before | after |
|---|---|---|
| worst_intra_pair_skew | 0.983 HOLDS | **1.353 BROKEN** (bound 1.100) |
| min_pair_coupled_fraction | 0.777 | 0.582 HOLDS |
| vias_per_si_net | 2.727 | 2.474 HOLDS |
| **verdict** | **ALL HOLD** | **3 of 4** |

The "40-net subset — receipt Pass" headline is **withdrawn**. It also failed to reproduce on merged main *before* this fix (1.347mm), so the original Pass was a property of one branch state, not a reproducible result.

| full board | before | after |
|---|---|---|
| worst_group_skew | 9.297 HOLDS | **8.397** HOLDS |
| worst_intra_pair_skew | 37.853 BROKEN | **38.521** BROKEN |
| min_pair_coupled_fraction | 0.000 BROKEN | **0.101** BROKEN |
| **vias_per_si_net** | **2.766 HOLDS** | **3.128 BROKEN** (bound 3.000) |
| **verdict** | **2 of 4** | **1 of 4** |

`vias_per_si_net` broke *as coverage improved* — the board routes more connections (0.994 vs 0.983) and pays in vias. Worth stating plainly rather than reporting only the routability gain.

### Short accounting

"The import carries ~906 pad-level shorts, and routing only 40 nets adds ~409" → the import carries **258**, and the 40-net route adds **27**. The conclusion is unchanged and still correct: `Short` is not route-attributable on this fixture; use `Clearance`.

### Wall-clock — the `<30 min` target is met

| stage | before | now |
|---|---|---|
| route | 113 min (pre-GPU-fix) | **958.9 s ≈ 16.0 min** |
| full chain | 2 h 51 m → 67 min | **1,634 s ≈ 27.2 min** |

Routability **0.994** is also a new record (prior best 0.985).

### Incidental gain

Pair-first now couples **47** pairs in round 0, up from 43 — overlapping pads had been sealing the BGA fields that pairs escape through.

## What did NOT move

**No receipt bound was adjusted.** The human board's anchor re-measures **bit-identical**: `9.756 / 1.074 / 0.857 / 2.265`, ALL HOLD. Pad angles feed pad geometry, not trace lengths, so the calibration anchor and the envelope argument it supports survive intact.

**The claim that survives with a corrected number**: the finishing pass still introduces no clearance violations. Published as "identical with the pass on and off (74)"; corrected, the 40-net board scores `Clearance` **23** — exactly the floor.

## Reviewer notes

- Body text in `gpu-router-m7-pair-si.md` is left as originally published, with a correction section appended, so the record shows what moved rather than silently overwriting it.
- `~/Downloads/cm5-vcad-fab/FAB_NOTES.md` is corrected in place (outside the repo, so not in this diff). The **gerbers do not need regenerating**: `board.kicad_pcb` round-trips identically to the fixture under the corrected reader (1,403 pads across 227 rotated footprints, zero mismatch) — the old writer emitted the raw absolute angle verbatim, so the copper was always correct. Only the numbers computed from it were wrong. Its `drc_report.txt` still carries the old `Clearance 74` and is flagged as predating the correction.
- Memory files `cm5-fab-ready-campaign.md` and `cm5-si-campaign.md` updated with the corrected figures and the lesson: when an importer changes, re-derive the baseline before trusting any delta.
- This PR depends on #684; it is merged into this branch.


ecto:
## Route spread — second run complete

Second full-board route on the corrected importer:

| | run 1 | run 2 |
|---|---|---|
| routability | 0.994 | **0.994** |
| board | 3,791 seg / 895 vias | **identical** |
| route wall-clock | 958.9 s | **895.5 s** |

The route is deterministic given the tree — both runs produced a byte-identical board. Only wall-clock varies (~7%, from machine contention).

**One caveat, stated rather than buried:** the full-chain figure (27.2 min) rests on **one** complete run. Run 2's route stage finished normally, but its `si_finish` stage was truncated when the machine's disk filled, so its 1,288 s is not a valid chain sample and is not averaged in. The route-stage spread above is from two complete runs; the chain number needs a second clean run to earn the same standing.

Both runs also shared the machine with another session's full-board route (load average 23.4), so the timings are pessimistic. The `<30 min` conclusion holds a fortiori.

All DRC and SI figures in this PR were already confirmed by two identical runs each and are unchanged.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

PR #699, opened 2026-07-25
fab-prep: probe and commit the whole via barrel, drill included

ecto:
The CM5 fab package failed closed with exactly 8 route-attributable violations. Both families come from one place: `vcad-ecad-fabprep`'s window-router committer, the only via emitter on the board whose barrels the legality oracle never saw. KiCad's independent DRC confirms all of them at the same coordinates (0 on the stripped baseline).

## What was wrong

**Barrel over a foreign trace — 2 defects, each raising a Short and a Clearance.** [`commit_path`](crates/vcad-ecad-fabprep/src/verdict.rs) committed the via annulus on the two *endpoint* layers only, and the probe-then-commit gate above it probed segments and never a via at all. So a barrel was committed blind, and its interior layers stayed invisible to the next path in the same round. Both defects are `FCu..In2Cu` barrels with the victim trace on `In1Cu` — the single interior layer, at 0.000mm.

**Via-to-via hole-to-hole — 4 defects.** The committer never called `commit_drill`, the gate never called `probe_via_drill`, and fab-prep called `route_window_complete` rather than `_pinned`, so `via: None` left the search's `drill_floor` at zero and `barrel_ok` a no-op. Hole-to-hole was enforced nowhere on this path. All four sit in the window *between* the two rules — centre spacing 0.29–0.36mm, i.e. copper-legal (≥ 0.21 + 0.08) and drill-illegal (< 0.12 + 0.25). Three of the four are two single-ended nets, so this is not specific to the pair router; `/PCIe.RX_P`/`_N` are twins only incidentally (different windows).

The router proper is clean here — every `route_all` commit goes through `validate_and_commit`, which does probe `span_slice` and `probe_via_drill`. `examples/cm5_verdict.rs`, which `verdict.rs`'s module docs claim it was lifted out of "so they don't drift", also already does all of this. It drifted.

## What changed

- `vias_legal`: drill probe, annulus on every spanned layer, and the path's own barrels against each other (the session cannot judge those until they are committed). Folded into the per-path gate, so a path with an illegal barrel downgrades itself to an honest unknown rather than being committed.
- `commit_path`: discs on **every** spanned layer, `commit_drill`, `path_vias` so a multi-step transition is one barrel instead of two stacked drills, and net-class via geometry instead of `default_rules`.
- The real `ViaClass` is handed to the search, so its grid pitch carries the hole-to-hole floor instead of proposing paths the gate must discard.
- `via_geom_for` promoted to `router::classes` as one source of truth (was private in `legalize`).

## Reviewer notes

- **Two fixture tests move to `max_rounds: 0`.** They set `route_remaining: false`, but that option gates only the *opening* verdict pass — the fix loop re-routes regardless. With the gate in place the loop now strips the crossing pair and honestly declines to re-route it into the same 0.3mm-drill congestion, which moves the blocker from geometry to connectivity. Giving the loop nothing to strip keeps the geometric offender under test, which is what those two assert.
- **This can make the connectivity number worse before it makes it better.** A path that used to commit illegally now stays unrouted, and the loop's objective is geometric only — a stripped net it cannot replace still scores as clean (the CM5 run reports 293 → 365 unconnected on arrival vs completion). Fail-closed is the right behaviour, but if convergence now blocks on connectivity rather than on the 8, the next lever is routability — terminal-layer pinning, which fab-prep also dropped versus `cm5_verdict` — not the gate.
- 5 new regression tests pin both bypasses (interior-layer span, drill vs existing hole, intra-path barrels, and that the commit indexes both). `cargo test -p vcad-ecad-fabprep` 36/36, `-p vcad-ecad-pcb` 297/297, clippy and fmt clean.
- **Full-board CM5 verification is still running** (a fresh `cm5_bench` for the arriving board, plus `fab-prep` over the packaged one). The mechanism above is established by code and by the measured geometry of all 6 offending vias; the before/after violation counts are not in yet, so this should not land on the strength of the CM5 numbers until they are posted here.

### DIFF
--- a/changelog/entries/2026-07-25-fabprep-via-commit-parity.json
+++ b/changelog/entries/2026-07-25-fabprep-via-commit-parity.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-25-fabprep-via-commit-parity",
+  "version": "0.9.4",
+  "date": "2026-07-25",
+  "category": "fix",
+  "title": "fab-prep no longer commits vias its oracle never saw",
+  "summary": "The fab-prep router committed via barrels unprobed and unindexed, landing inner-layer shorts and hole-to-hole violations; every barrel is now probed and committed on all layers it spans, drill included.",
+  "features": ["pcb", "routing", "drc", "fab"],
+  "mcpTools": ["fab_prep", "run_drc"]
+}
--- a/crates/vcad-ecad-fabprep/src/lib.rs
+++ b/crates/vcad-ecad-fabprep/src/lib.rs
@@ -628,8 +628,12 @@ mod tests {
             &FabPrepOptions {
                 route_remaining: false,
                 prune_dangling: false,
-                // One round is enough to prove the loop tried and stopped.
-                max_rounds: 1,
+                // No rounds: with nothing allowed to strip the crossing pair,
+                // the short survives and must be named. (Given a round, the
+                // loop strips both nets and the fail-closed commit gate
+                // refuses to re-route them into the same 0.3mm-drill
+                // congestion — an honest unroutable, tested separately.)
+                max_rounds: 0,
                 ..Default::default()
             },
         )
--- a/crates/vcad-ecad-fabprep/src/render.rs
+++ b/crates/vcad-ecad-fabprep/src/render.rs
@@ -445,7 +445,10 @@ mod tests {
             &FabPrepOptions {
                 route_remaining: false,
                 prune_dangling: false,
-                max_rounds: 1,
+                // No rounds: the strip-and-re-route loop must not be given a
+                // chance to replace the short with an honest unroutable, so
+                // the geometric offender is what the report has to name.
+                max_rounds: 0,
                 ..Default::default()
             },
         )
--- a/crates/vcad-ecad-fabprep/src/verdict.rs
+++ b/crates/vcad-ecad-fabprep/src/verdict.rs
@@ -10,10 +10,13 @@
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use vcad_ecad_pcb::ratsnest::{compute_ratsnest, NetConnection, Netlist, NetlistNet};
-use vcad_ecad_pcb::router::complete::{route_window_complete, CompleteOutcome};
+use vcad_ecad_pcb::router::classes::via_geom_for;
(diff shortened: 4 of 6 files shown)


ecto:
## CM5 verification: all 8 route-attributable violations gone

`vcad fab-prep --calibrate-rules` over the packaged CM5 board (1,921 traces / 363 vias — the artifact that carried the 8), before → after this change:

| rule | before | after |
|---|---|---|
| Clearance | 2 | **0** |
| Short | 2 | **0** |
| HoleToHole | 4 | **0** |
| **total route-attributable** | **8** | **0** |

`HoleToHole` no longer appears in the delta table at all (0 baseline / 0 final), and a full `drc_json` over the output board reports zero hole-to-hole violations. At each of the six defect sites, only one via now remains — the colliding partner is either legally re-placed or honestly unrouted:

```
(143.042, 74.535) -> /ETH.3_N only
(143.327, 74.706) -> none
(151.252, 84.546) -> none
(131.850, 98.485) -> /PCIe.RX_N only
(144.177, 68.458) -> none
(138.400, 87.450) -> /BCM2712 Misc IO/BT_UART.TX only
```

The opening verdict pass committed **118 routes with the fixed committer and produced zero route-attributable violations** — which is the direct evidence for the fix, since those 118 commits are exactly the population that used to mint them.

## It still does not converge — on connectivity, by 2 nets

Final connectivity is 365 on arrival → 367 on completion, a regression of **2**, and that is what now trips the guard. So the geometric claim is clean and the blocker moved, as predicted in the PR body. Whether -2 is acceptable is a judgement call for the campaign, not something this PR should decide.

Two things worth a reviewer's attention:

1. **The blocker string disagrees with the table.** The CLI printed "98 more nets are unconnected or islanded", while the report's own connectivity line says 2. The blocker text is captured when the loop gives up, but the loop then restores the best board it saw; the stored message is stale relative to the board actually written. Cosmetic, in a receipt whose whole job is to be exact — worth a follow-up, out of scope here.
2. **This input already contained the 8**, so reaching 0 required the loop to strip 19 nets in round 1 and the gate to refuse re-adding them illegally. The cleaner test — a fresh `cm5_bench` board fed to a fixed fab-prep, where the 8 should never be minted in the first place — is still running and I'll post it when it lands.

_🤖 Verified by [Claude Code](https://claude.com/claude-code)_

PR #703, opened 2026-07-25
fix(router): clear every same-net bypass and self-colliding via the router creates

ecto:
The CM5 fabrication package was blocked by 22 route-attributable violations
(21 `SameNetBypass`, 1 `HoleToHole`). This clears the `HoleToHole` and 14 of
the 21 bypasses, at unchanged routability and improved connectivity.

## Measured, full CM5 route (`--calibrate-rules --skip-routing --accept MinTraceWidth --accept Short`)

| | before | after |
|---|---|---|
| HoleToHole | 1 | **0** |
| SameNetBypass | 21 | **7** |
| Unconnected/islanded on arrival | 293 | **271** |
| UnconnectedNet (final) | 275 | **254** |
| routability | 0.988 | **0.988** |

No net was demoted to reach zero `HoleToHole` — connectivity improved rather
than being traded away.

## Four defects, each found by measurement

**The legalization loop could only ever repair four violations.** It re-judged
the whole board after every single fix under a `MAX_FIX_ROUNDS = 4` cap, so the
pass spent its entire budget on the first four: 21 bypasses in, 17 out. A round
now clears every bypass it can prove independent — independence is per net,
since a same-net contact's hop count is measured along that net's own adjacency
graph — and the cap bounds unmasking, not violations.

**Legalization judged a board that was not the one shipping.** The router hands
over ~1855 dead traces the caller prunes moments later, and deleting dead copper
*lengthens* the surviving conductor chains: a contact 5 hops apart through
debris is 9 hops apart once it is gone. Legalization now prunes at the top of
each round.

**A path's vias were never checked against each other.** Every via in a
candidate is probed before any is committed, so each is judged against a session
that does not yet contain its siblings. A descent stepping In5→In7 then In7→BCu
emits exactly that shape — two same-net barrels a fraction of a millimetre
apart, each legal alone. That is how the CM5's `HoleToHole` (0.116mm < 0.250mm
at (132.000, 95.345), a `/MIPI0.D0_N` stack) reached the drill file. Siblings
are now checked explicitly; a stacked pair landing on one point folds into a
single barrel via `span_union` rather than being rejected, since every layer of
the union was already probed.

**`si_finish` had no legalization at all.** It rips, re-routes and prunes after
`route_all` is done (483 dead traces on the CM5) and nothing judged the result.
Its pair-polish gate counted only `Clearance`/`Short` *errors*, so `SameNetBypass`
(a warning) sailed through; its meander gate probes each segment against a board
with that net's own copper removed, so a serpentine folding back on its own run
reads as legal. Both gates now count bypasses, non-regressively — declining an
optimization costs nothing, the pair keeps the copper it had.
`repair_same_net_bypass` then clears what remains, board-level.

## Not cleared by deletion

Every repair is gated on the continuity oracle: a segment goes only when its net
ends up with no more islands and no less coverage. Bypasses whose two sides are
*both* load-bearing are left standing rather than cleared by deleting
connectivity — the trap `fab-prep`'s arrival-vs-completion guard exists to catch.
`board_repair_refuses_to_clean_by_deletion` pins that behaviour.

## What still blocks the package

7 bypasses remain, on 4 nets (`/USB3-0.RX_P` ×3, `/ETH.0_N` ×2, `/MIPI0.D1_P`,
`Net-(C259-Pad2)`), identical across two independent routes — so they come from
detailed routing, not from `si_finish`. They are one class: a via or leg of one
connection landing on another part of the same net's copper mid-body, both sides
load-bearing. `/ETH.0_N` is representative — a connector via sits exactly on its
own leg's centerline and a 0.08mm neck leaves it at a different angle.

`fab-prep`'s own fix loop confirms the diagnosis independently: it *can* reach a
geometrically clean board, but only by stripping 51 nets, and rejects that.

Preventing them needs the search to stop treating its own net's copper as a
free corridor — same-net copper is invisible to the probe, which is what lets a
connector land mid-leg — and join only at intended points. That is a real change
to the routing search with its own routability risk, and it is not landed here:
one speculative fix was written, measured against the actual geometry, found to
address a different shape than the one occurring, and reverted rather than
shipped unvalidated.

## Verification

- `cargo test -p vcad-ecad-pcb` — 302 passed
- `cargo clippy -p vcad-ecad-pcb --features gpu -- -D warnings` — clean
- `cargo fmt -p vcad-ecad-pcb` — clean

New regression tests: `every_independent_bypass_is_cleared_not_just_the_first_few`
(7 nets, more than the old 4-repair cap could ever clear),
`bypass_candidates_found_when_the_contact_point_is_off_centerline`,
`board_repair_clears_a_redundant_bypass`,
`board_repair_refuses_to_clean_by_deletion`,
`stacked_via_spans_union_into_one_barrel`. Two `#[ignore]`d fixture harnesses
(`VCAD_LEGALIZE_FIXTURE`) reproduce the CM5 case in seconds instead of a 25-minute route.

### DIFF
--- a/changelog/entries/2026-07-25-router-clears-same-net-bypasses.json
+++ b/changelog/entries/2026-07-25-router-clears-same-net-bypasses.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-25-router-clears-same-net-bypasses",
+  "version": "0.9.4",
+  "date": "2026-07-25",
+  "category": "fix",
+  "title": "Autorouter clears every same-net bypass it creates",
+  "summary": "Post-route legalization repaired at most four violations per run and judged the board before dead copper was pruned; a path's own vias were never checked against each other.",
+  "features": ["pcb", "routing", "drc", "fab"],
+  "mcpTools": ["route_nets", "fab_prep", "run_drc"]
+}
--- a/crates/vcad-ecad-pcb/src/router/auto.rs
+++ b/crates/vcad-ecad-pcb/src/router/auto.rs
@@ -773,7 +773,17 @@ pub fn route_all_with_opts(
     // autoroute pass never returns HoleToHole or SameNetBypass violations of
     // its own making.
     let stitch_via_pts: Vec<Vec2> = stitch.vias.iter().map(|(p, _)| *p).collect();
-    let legal = super::legalize::legalize(pcb, &mut traces, &mut vias, &stitch_via_pts);
+    // Snapshot before legalization, not before the prune below: legalization
+    // now prunes dead copper itself (it has to — see `legalize`), so a net that
+    // ends up with nothing left has to be detectable across both passes.
+    let anchors_before = net_copper_anchors(&traces, &vias);
+    let legal = super::legalize::legalize(
+        pcb,
+        &mut traces,
+        &mut vias,
+        &stitch_via_pts,
+        opts.prune_dangling_copper,
+    );
     if legal.merged_vias > 0
         || legal.pruned_traces > 0
         || !legal.demoted.is_empty()
@@ -802,7 +812,6 @@ pub fn route_all_with_opts(
     // the function, so there is no session state left to keep in step.
     let mut pruned_out: Vec<(String, Vec2)> = Vec::new();
     if opts.prune_dangling_copper {
-        let before = net_copper_anchors(&traces, &vias);
         let (dead_traces, dead_vias) = super::legalize::prune_dangling(pcb, &mut traces, &mut vias);
         if dead_traces > 0 || dead_vias > 0 {
             log::info!(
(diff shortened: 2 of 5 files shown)


ecto:
Second commit on this branch (`462a1157`) is a reporting fix, and it turned up something bigger than the bypasses.

**The connectivity oracle is mismeasuring the board.** Run against the *human-routed* CM5 — the shipping Raspberry Pi board — it reads **115/374 signal nets connected (0.307)**, worse than our router's 0.399. A known-good board cannot be 30% connected, and 226 of those nets are short by exactly one connection: a systematic gap, not 226 accidents.

Measured cause, over the fixture's 958 netted pads on rotated footprints:

| convention in `geometry::pad_world_position` | pad centre lands on a trace endpoint | …endpoint carries the pad's own net |
|---|---|---|
| current (`+fp.rotation`) | 424 | **159** |
| negated (`−fp.rotation`) | 472 | **471** |

KiCad's `at` angle is CCW as drawn, in a Y-down frame that `kicad_pcb::read_at` imports verbatim, so pad offsets need the opposite sign. A 2-pad passive at −90° comes out with its pads swapped and each attributed to its neighbour's net — e.g. `C6` on the CM5: pad 1 (`Net-(C5-Pad2)`) is placed where pad 2 (`GND`) belongs.

**Not fixed here on purpose.** It is a pipeline-wide convention touching Gerber, pick-and-place, DRC and the router, it interacts with the pad-rotation work in #688, and `vcad-ecad-invariants` exists precisely to pin a pad's world rect across all of them. It wants its own change with that harness, not a sign flip smuggled in beside a reporting fix.

The practical consequence for this PR: the `UnconnectedNet` column in the fab-prep table — and the "271 unconnected on arrival" guard — are inflated by this, on both the baseline and the final board. The *route-attributable* delta is unaffected, since both sides are measured the same way.

New: `drc::net_pad_groups` (agrees with `UnconnectedNet` net-for-net, test-pinned), `SiFinishReport::emptied_nets`, and a `net_census` example that prints the scoreboard for any board.

PR #712, opened 2026-07-26
fix(motor): calc_motor sees saturating stator teeth

ecto:
## The bug

`calc_motor` (PM mode) computes air-gap flux via a magnetic-equivalent-circuit model that treats iron as **linear**. It therefore over-predicts `Kt` on any machine whose stator teeth saturate, with no warning. It also had no tooth concept at all, so it structurally *could not* detect the case.

Found while designing a 24-slot / 20-pole axial-flux QDD actuator (⌀76→⌀120 annulus, 4 mm N42, 1.0 mm gap, `iron_mu_rel: 4000`):

- tooth pitch at mean radius 49 mm = 2π·49/24 = **12.83 mm**
- at 50% tooth width (6.4 mm), flux concentrates **2.0×**
- `B_gap` reads a comfortable **1.03 T** → `B_tooth ≈ 2.06 T`, well past the ~1.5–1.7 T knee for M19-class silicon steel

Nothing in the output hinted at it. Someone sizing an actuator off that `Kt` builds the wrong motor.

## What changed

**Kernel — `crates/vcad-ecad-sim/src/airgap.rs`**

- `TeethSpec { slots, tooth_width_mm, mean_radius_mm, tooth_path_mm }` with `tooth_pitch_mm()` and `concentration()` (= pitch/width, clamped ≥ 1 so a wide tooth can never *dilute*).
- `iron_js_t: Option<f64>` on `AirGapSpec` switches the soft iron to the arctangent B–H law, reusing `vcad_kernel_em::material::{Saturation, nu_from_b}` — the same law the axisymmetric FV solver uses, not a second one. New dep: `vcad-ecad-sim` → `vcad-kernel-em`.
- New `airgap_solve() -> AirGapSolution`: gap/tooth/yoke flux density, concentration factor, `nonlinear`, `converged`, and `warnings`. `airgap_flux_density()` is now a thin wrapper over it.
- Warnings fire past the knee (`0.85·J_s`, else `SILICON_STEEL_KNEE_T = 1.5`) for **both tooth and yoke**, including on a *linear* solve. Announcing the optimism is the whole point — the old `note` ("no slotting/saturation/losses") was too easy to miss when the number looked great.

**Bisection, not Picard.** The plan called for a Picard loop; successive substitution on ν oscillates here because dν/dΦ is steep past the knee, and convergence depended on a hand-picked damping factor (relax 0.7 and 0.05 both failed to converge on the thin-tooth case; 0.1 and 0.2 passed). That is exactly the kind of hidden knob that makes a saturating result untrustworthy. Instead the saturating branch bisects the MMF balance

```
F = Φ·(R_m + R_g) + Σ ν(Φ/A)·(l/A)·Φ,    F = φ_r·R_m
```

whose RHS is strictly increasing in Φ, bracketed by `[0, linear-iron flux]`. 80 halvings exhaust f64, no tuning constant, always converged. (`axisym.rs` keeps its Picard loop — it solves a field, not a scalar.)

**MCP — `calc_motor`**

- New `magnet` fields: `slots` + (`tooth_width_mm` | `tooth_fraction`), optional `mean_radius_mm` (defaults to the midpoint of the radii already passed), `tooth_path_mm`, `iron_js_t`.
- Reports `tooth_flux_tesla`, `tooth_concentration`, `yoke_flux_tesla`, `iron_model`, `warnings`.
- Emits a `tooth_flux_density` claim (`mec-tooth-concentration` / `mec-saturating-iron`) so an FEA pass grades the tooth field directly.
- The `note` now says the iron is LINEAR and `Kt` is OPTIMISTIC, and names the flag that fixes it.
- Tooth field comes from the **raw** gap flux: fringing redistributes flux under the pole, it doesn't remove it, so the tooth carries the same flux either way. Stated in `tooth_note`.
- Falls back to the scalar WASM binding with a TS-mirrored (pure-geometry) tooth calc, so the warning survives a kernel that predates `ecadAirgapSolve`.

`tooth_path_mm = 0` (the default) makes the teeth diagnostic-only — reported and gated, but adding no reluctance.

## Backward compatibility

Saturation is opt-in. With no `iron_js_t` and no `teeth`, every existing caller gets bit-identical numbers — asserted by `saturation_is_opt_in_and_changes_nothing_by_default`.

## Testing

Five new Rust cases in `airgap.rs`:
- `tooth_pitch_and_concentration_match_the_geometry` — 12.83 mm pitch, k ≈ 2.00, and a full-pitch tooth never dilutes
- `linear_iron_hides_a_saturating_tooth_and_says_so` — the motivating QDD case: B_gap ~1.0 T, tooth past the knee, warning present
- `saturating_iron_is_materially_pessimistic_vs_linear` — 4 mm teeth (k ≈ 3.2) in the loop: the saturating solve lands >5% below the linear one
- `below_the_knee_linear_and_saturating_agree` — wide teeth, ferrite-class drive: the two models agree to <3% and no warning fires
- `yoke_saturation_also_warns` — thin back-iron, no teeth at all

Five new MCP cases covering the tooth report, `tooth_width_mm` ↔ `tooth_fraction` agreement, the saturating solve dropping `Kt`, silence when no tooth geometry is given, and input rejection.

- `cargo test -p vcad-ecad-sim -p vcad-ecad-diff` — green
- `cargo clippy -p vcad-ecad-sim -p vcad-ecad-diff` + `cargo fmt --all --check` — clean
- Full MCP suite: 985 passed, 3 skipped
- `tool-surface.fixture.json` regenerated (`UPDATE_TOOL_SURFACE=1`)
- Kernel WASM rebuilt locally to exercise the real `ecadAirgapSolve` path, then reverted to `origin/main` per the single-writer rule

## Not done

- **The `simulate_em` cross-check.** `problem: "planar_magnetostatics"` with `js_t` on the iron regions is the right in-repo oracle — unroll the machine at mean radius into a linear periodic cross-section (for 24s/20p the periodic unit is 10 poles / 12 slots, since gcd(24,20)=4 and `periodic_x` needs an even pole count). That's a separate piece of work from the model change.
- `cargo clippy --workspace --all-targets` is red on `vcad-kernel-gpu` and `vcad-kernel-booleans`; both fail identically on a clean tree.


ecto:
No change made — the finding is describing the PR working as intended.

`airgapSolve`, `AirGapSolutionResult`, and `TeethSpecInput` **are** added and exported in this PR (`packages/engine/src/index.ts:219,276-277`, in commit `f4de1a9`). The observation that they are absent from `main` is just the pre-merge state; that is what the diff changes.

There is no split-release risk here: `@vcad/engine` and `@vcad/mcp` are workspace packages in the same repo, built together, and `mcp-publish.yml` ships a single self-contained tarball on each main merge — the MCP server cannot be published against a stale engine. Verified locally: `tsc --noEmit` on `packages/mcp` is clean and the full MCP suite (985 passed) exercises `calcMotor` through the real `airgapSolve` path.

PR #713, opened 2026-07-26
fix(em): simulate_em can solve a strongly saturating magnetostatics problem

ecto:
## The problem

`simulate_em` with `problem: "planar_magnetostatics"` and `js_t` on iron regions failed on any device saturated hard enough to need more than 80 Picard iterations:

```
Error: Picard not converged after 80 iterations (largest reluctivity update 1.346e-3)
```

Passing `tol: 1e-6` and `max_sweeps: 400000` changed **nothing** — the residual came back bit-identical. Those govern the inner SOR sweep; the outer nonlinear loop's options were always `PicardOptions::default()` and were exposed nowhere. Strongly saturated machines are a primary use case for this solver, so the tool was simply unusable for them.

## What the measurement showed

The reported diagnosis was oscillation wanting deeper under-relaxation. It isn't. On the repro (unrolled axial-flux section, saturable teeth past the knee) the loop **contracts steadily** at the stock relaxation — 0.92 per iteration — and just needs ~140 iterations against a cap of 80.

Lowering `relax`, the intuitive fix, is strictly worse:

| `relax` | iterations to converge |
|---|---|
| 0.7 | 137 |
| 0.5 | 189 |
| 0.3 | 309 |
| 0.15 | no convergence in 400 |

That reframed the fix: the cap was the bug, and any error message advising under-relaxation would have sent callers the wrong way.

## Changes

- **Knobs plumbed through** — `picard_max_iters`, `picard_tol`, `picard_relax`, `picard_adaptive` on `EmSimOptions` and the `simulate_em` schema, validated fail-closed (`picard_relax: 1.5` errors rather than silently clamping).
- **Default cap 80 → 300.** Raising a cap can only convert failures into answers; it never changes a solve that already converged. The repro now converges **on the defaults, no options**, in 140 iterations.
- **The error names the knob, measured not guessed.** The loop tracks its residual's contraction rate over a trailing window and branches on it:
  - still falling → *"The residual IS still falling, at 0.9234 per iteration — the iteration cap arrived first. Retry with `picard_max_iters: 128`. Do not lower `picard_relax` here."* (the suggested value is extrapolated from the observed decay)
  - stalled → name `picard_relax`, and note the device itself may be over-driven
  - too few iterations to tell → say exactly that, rather than diagnosing from nothing

  Every branch states up front that `tol`/`max_sweeps` are the SOR knobs and do nothing here.
- **`PicardReport` survives failure** — `SolveError::picard_report()`, carried as JSON alongside the message, so a caller can judge how far off the material state was instead of parsing prose.
- **Latent correctness fix:** the residual was measured on the *damped* update, so lowering `relax` would have shrunk the reported residual and faked convergence. It is now the undamped fixed-point delta.

## On adaptive under-relaxation

Implemented, but **ships off by default**, with the evidence in the doc comment. On this device it cost 1.7× the iterations (236 vs 140) by reacting to the ordinary residual jitter of a healthy solve, and starting from `relax=0.3` it ratcheted to the floor and turned a slow-but-converging run into a failure. It's the right lever for a genuine limit cycle — which this fixture is not — so it's an opt-in knob rather than a default-on heuristic I can't demonstrate a win for.

Aitken/Anderson acceleration and a Newton step on the B–H law are **not** in this PR. They're worth doing, but they should be justified against a device that genuinely limit-cycles.

## Testing

New regression tests in `planar.rs`: the saturating section (6.4 mm teeth, ~1.8 T) on defaults and with a raised cap; an 8 mm control below the knee, so the failure boundary is explicit rather than implicit; the slow-contraction diagnosis; the fail-closed report contract; and a deterministic `PicardDamping` unit test separating jitter from stagnation.

52 `vcad-kernel-em` tests + 9 validation tests + 980 MCP tests pass. Clippy and fmt clean. Tool-surface fixture regenerated.

Verified end to end through a locally rebuilt kernel WASM against the actual `simulate_em` path; the generated artifacts were reverted per CLAUDE.md and will be refreshed by `wasm-refresh.yml` on merge.

## Reviewer notes

- The `adaptive` default and the raised `max_iters` default are the two judgement calls here — both are argued from numbers in the code comments, and both are easy to flip if you read the tradeoff differently.
- Does not touch `calc_motor`'s missing saturation (task_fdde42af). `simulate_em` can now serve as its oracle, which it couldn't before.

### DIFF
--- a/changelog/entries/2026-07-25-em-saturated-picard.json
+++ b/changelog/entries/2026-07-25-em-saturated-picard.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-25-em-saturated-picard",
+  "version": "0.9.4",
+  "date": "2026-07-25",
+  "category": "fix",
+  "title": "simulate_em solves strongly saturating magnetostatics",
+  "summary": "Saturated motor cross-sections no longer fail with an unfixable 'Picard not converged': the nonlinear loop's cap, tolerance and relaxation are now caller-settable, and the default cap was raised.",
+  "features": ["em", "magnetostatics", "motors"],
+  "mcpTools": ["simulate_em"]
+}
--- a/crates/vcad-kernel-em/src/axisym.rs
+++ b/crates/vcad-kernel-em/src/axisym.rs
@@ -114,23 +114,150 @@ impl Material {
 }
 
 /// Options for the Picard (successive-substitution) nonlinear loop.
+///
+/// These govern the **outer** nonlinear loop only. The inner SOR sweep
+/// has its own [`SolveOptions::tol`] / `max_sweeps`; loosening those does
+/// nothing for a device whose materials refuse to settle.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct PicardOptions {
     /// Hard cap on outer iterations.
     pub max_iters: usize,
     /// Convergence: largest relative change of any cell's reluctivity.
     pub tol: f64,
-    /// Under-relaxation on the reluctivity update, in (0, 1].
+    /// Under-relaxation on the reluctivity update, in (0, 1]. With
+    /// `adaptive` set this is the *starting* (and maximum) value.
     pub relax: f64,
+    /// Back `relax` off when the residual makes no net progress over a
+    /// 10-iteration window, and give it back when contraction resumes.
+    ///
+    /// **Off by default, and that is a measured choice, not caution.** On
+    /// the saturating motor section in this crate's tests it cost ~1.7x
+    /// the iterations (236 vs 140) by reacting to the residual jitter of
+    /// a healthy solve, and on the same device started from `relax = 0.3`
+    /// it ratcheted to the floor and turned a slow-but-converging run
+    /// into a failure. Damping is the right lever for a genuine limit
+    /// cycle; on this solver's B–H fixed point, slow contraction is the
+    /// far more common failure and wants `max_iters` instead. Turn this
+    /// on when a device is demonstrably oscillating — the failure message
+    /// says which mode you are in.
+    pub adaptive: bool,
 }
 
 impl Default for PicardOptions {
     fn default() -> Self {
         Self {
-            max_iters: 80,
+            // 80 was too tight to be a default: a saturating motor
+            // cross-section contracts steadily at this relaxation and
+            // simply needs ~140 iterations, so the old cap turned an
+            // ordinary machine into a hard failure with no caller-side
+            // recourse. Raising a cap can only convert failures into
+            // answers — it never changes a solve that already converged.
+            max_iters: 300,
             tol: 1e-4,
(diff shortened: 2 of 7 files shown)


ecto:
@chojiai The Correctness finding is a false positive — the axisym driver **was** upgraded in this PR. The finding's own text says it was *"confirmed on `main`"*, which is the ref it read; the change is on `f4767d5`:

```
$ git show f4767d59:crates/vcad-kernel-em/src/axisym.rs | grep -n 'PicardDamping::new|damp.relax()|damp.observe|max_db.max|decay_rate: damp'
592:        let mut damp = PicardDamping::new(popts);
621:            let relax = damp.relax();
643:                    max_db = max_db.max(delta.abs());   # undamped residual
648:            damp.observe(rel);
668:            relax: damp.relax(),
670:            decay_rate: damp.decay_rate(),
```

Both drivers got the identical treatment, and both WASM call sites (`axisym_magnetostatics` and `planar_magnetostatics`) were rewired from `PicardOptions::default()` to the caller-settable `em_picard_options(&opts)?`. `PicardOptions` itself is *defined* in `axisym.rs` and shared by both drivers, so the raised `max_iters: 300` default (line 155) and the new `adaptive` field (line 158) apply to axisym by construction — the file the finding points at is the one that defines them.

One fair observation underneath it: the new *regression* fixtures are planar-only, because the reported repro is a planar motor cross-section. Axisym's nonlinear path is covered by the existing `nonlinear_solenoid_matches_the_exact_bh_law`, `nonlinear_with_huge_saturation_matches_linear`, and `secant_inductance_droops_into_saturation` tests, which pass against the new residual measure — that last point matters, since switching to the undamped residual made those assertions strictly tighter.

No code change made.

ecto:
CI status: **Rust (stable) and Rust (nightly) pass** — those are the jobs that exercise this change (52 `vcad-kernel-em` unit tests + 9 validation tests, clippy, fmt). WASM build, torture track, KiCad export, and the no-committed-artifacts guard all pass too.

**TypeScript fails, but not because of this PR.** The two failures are in `packages/mcp/src/__tests__/session-restart.test.ts` — session-id entropy and a path-traversal guard in the file session store — which this branch doesn't touch. The same two tests fail on `main`'s tip commit b7938593 ([run 30180165535](https://github.com/ecto/vcad/actions/runs/30180165535), also `Failed Tests 2`, same file):

```
FAIL session-restart.test.ts > boot-generation tagging > ids stay unguessable
FAIL session-restart.test.ts > FileSessionStore ... > refuses an id that could escape the session directory
```

The first also looks flaky rather than merely broken — it asserts a base64url length `>= 16` and reported `14` on main vs `7` here, i.e. a varying value across runs. Worth its own fix; I didn't fold it into this PR since it's unrelated and the traversal one may be a real security-relevant bug deserving its own review rather than a drive-by.

PR #714, opened 2026-07-26
fix(mcp): unbreak two session-restart tests failing on Linux CI

ecto:
The `TypeScript` CI job is red on `main`, and therefore on every PR branch. Two tests in `packages/mcp/src/__tests__/session-restart.test.ts` fail on Linux but pass on macOS. Neither is a bug in the code under test — both assertions were checking something other than the invariant they meant to pin.

Test-only change. No production behavior is modified.

## 1. `refuses an id that could escape the session directory`

The test probed an absolute host path:

```ts
expect(existsSync(join(dir, "..", "..", "etc", "passwd"))).toBe(false);
```

`dir` is `mkdtempSync(join(tmpdir(), "vcad-sessions-"))`. On Linux CI that's `/tmp/vcad-sessions-XXXX`, so the expression resolves to the **real `/etc/passwd`**, which exists — the assertion failed regardless of whether `FileSessionStore` was secure. macOS passed only incidentally: its tmpdir (`/var/folders/rs/.../T/...`) is deep enough that the same expression lands on a nonexistent path.

Fix: nest the session dir inside an outer sandbox and assert on the *sandbox's* contents. Nothing escaped iff the sandbox still holds only `sessions/` and `sessions/` is empty. The input set is also widened to `["../../etc/passwd", "../escaped", "..", "/abs/path"]` — `"../escaped"` is the one that would land *inside* the sandbox, so the assertion now observes an actual escape rather than the host filesystem.

**Teeth verified:** deleting the `/^[A-Za-z0-9_-]{1,128}$/` guard from `FileSessionStore.pathFor` makes the test fail (`expected { nodes: {}, roots: [] } to be null`). Restored afterward.

## 2. `ids stay unguessable — the token is a prefix, not a replacement`

This one was reported as possibly flaky. It is deterministic, and it's a real assertion bug:

```ts
expect(a.split("_")[2].length).toBeGreaterThanOrEqual(16);
```

Session ids are `doc_<n>_<boot><random>`, and the suffix is **base64URL — which uses `_` as a character**. When any of the 16 suffix chars is an underscore, `split("_")[2]` is truncated and the assertion fails. Measured failure rate: **22.4% per id**, and the test mints two, so ~1 CI run in 5. PR #712 was simply lucky.

Fix: strip the prefix with `id.replace(/^doc_\d+_/, "")` instead of splitting on a character that occurs in the payload.

## Verification

- `npm test -w @vcad/mcp` → **90 files, 980 passed / 3 skipped**
- Re-run under `TMPDIR=/tmp` to reproduce the shallow-tmpdir condition that breaks Linux CI → still green
- Traversal test confirmed to fail with the path guard removed

## Note for reviewers

A fresh worktree needs `npm ci` **and** `npm run build:live` before vitest can even collect this file — `src/live-html.generated.js` is generated and not checked in, and its absence surfaces as an unrelated-looking module-resolution error in `live-route.ts`.

### DIFF
--- a/packages/mcp/src/__tests__/session-restart.test.ts
+++ b/packages/mcp/src/__tests__/session-restart.test.ts
@@ -77,7 +77,7 @@ describe("boot-generation tagging", () => {
     // splitting on "_" and taking [2] truncates it ~22% of the time.
     const suffix = (id: string) => id.replace(/^doc_\d+_/, "");
     expect(suffix(a)).not.toBe(a);
-    expect(suffix(a).length).toBeGreaterThanOrEqual(16);
+    for (const id of [a, b]) expect(suffix(id).length).toBeGreaterThanOrEqual(16);
   });
 });
 
--- a/packages/mcp/src/__tests__/session-restart.test.ts
+++ b/packages/mcp/src/__tests__/session-restart.test.ts
@@ -13,7 +13,7 @@
  *  3. a mint under a non-durable store says so, up front.
  */
 import { describe, it, expect, beforeEach, afterEach, beforeAll } from "vitest";
-import { mkdtempSync, rmSync, existsSync } from "node:fs";
+import { mkdtempSync, mkdirSync, readdirSync, rmSync, existsSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { Engine } from "@vcad/engine";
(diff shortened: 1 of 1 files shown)


ecto:
Addressed the minor finding in 2848cf79: the suffix-length assertion now loops over both `a` and `b` rather than checking only `a`, so a truncation regression is caught whichever id trips it. `npm test -w @vcad/mcp` still green (19/19 in this file, run under `TMPDIR=/tmp`).

PR #730, opened 2026-07-26
Ray-trace triangle meshes so mesh-backed parts stop vanishing from renders

ecto:
## The bug

Both ray-traced paths skipped any solid without an analytic BRep:

```rust
let Some(brep) = s.solid.as_brep() else { continue; };
```

…and then errored only if **nothing** traceable remained. So a document mixing a BRep part with a mesh part rendered the BRep and silently omitted the mesh — no warning, no indication anything was missing. This hits real workflows: frozen `topology_optimize` results and imported STL/GLB parts are meshes.

## Kernel (`vcad-kernel-raytrace`)

- **New `intersect/triangle.rs`** — Möller–Trumbore. Double-sided (mesh winding isn't trustworthy on decimated/imported geometry, and the renderer face-forwards the shading normal anyway, so culling would only punch holes in parts). The determinant guard is size-relative, not absolute — an absolute epsilon rejects legitimate hits on µm-scale triangles.
- **Dispatch** — as the brief suspected, triangles need a different entry point: `intersect_surface` dispatches on `SurfaceKind` via downcasts and a triangle has no `SurfaceKind`. They're called straight from the BVH's mesh leaves instead.
- **`Bvh` builds over triangles too** — leaves now carry `prims: Vec<u32>`, and the `Bvh` owns a `BvhGeom` mapping those indices to either face IDs (`BRep { brep, faces }`) or triangles (`Mesh`). `Bvh::build_mesh(&TriangleMesh)` therefore reuses the SAH builder unchanged, so **the analytic path's behaviour and performance are untouched** — same code, same traversal. Degenerate triangles (repeated/out-of-range indices, zero area) are dropped at build time rather than wasting intersection tests.
- **Smooth shading** — mesh hits interpolate vertex normals barycentrically, falling back to the geometric normal when the mesh has none or the blend cancels.
- **API changes** — `RayHit` gains `tri: Option<u32>` (`RayHit::new` keeps its signature; new `RayHit::triangle` constructor sets a null `face_id`). `Bvh::brep()` and `CpuRenderer::brep()` now return `Option<&BRepSolid>`. `flatten()` stays BRep-only and returns empty for a mesh BVH — the GPU pipeline traces analytic surfaces and must not be handed face IDs that don't exist.

## Renderer (`vcad-render`)

- `rasterize_rt` builds a mesh BVH for non-BRep solids, crease-baked via `render_bake` (what `Solid::to_mesh`'s own docs prescribe for tracer-bound meshes).
- **`evaluate_vcad` now carries `ImportedMesh` roots through as mesh-backed solids.** Worth a close look — this is the one change beyond the brief's letter. Those roots evaluate to `EvaluatedPart { solid: None }`, so they were being dropped a level *above* the raytrace skip; without this, the two workflows the brief names (frozen topopt, dropped-in STL/GLB) still would not render. Side effect: they now also reach the **tessellated** path, where they were equally invisible before.
- **The drop path is honest** — a part with no traceable geometry is named in the error rather than silently omitted. This fails closed: a missing part in a render reads as a design that doesn't have it, which is worse than no render.

## Tests

The new render test asserts **coverage**, not absence of errors — a cube and a mesh box 40mm apart must produce two disjoint covered column runs with background between them. Under the old code only the cube survived, which frames to a single run filling the canvas, so the test fails against the bug. A test that merely checked "it did not error" would have passed.

Also covers: mesh vs. analytic cube entry/exit parity, vertex-normal vs. geometric-normal selection, degenerate-triangle rejection, empty meshes, mesh-flattens-to-nothing, and BRep `flatten()` still round-tripping face IDs.

## Verification

```
cargo test -p vcad-kernel-raytrace -p vcad-render   # green (68 + 68 lib, + integration)
cargo clippy -p vcad-kernel-raytrace -p vcad-render --all-targets -- -D warnings   # clean
cargo fmt   # clean
```

`cargo check -p vcad-kernel-raytrace --features gpu` and the downstream `vcad-kernel-dfm` / `vcad-ffi` consumers also compile.

Note for reviewers reproducing locally: `vcad-kernel-text` `include_bytes!`s a font out of `node_modules/`, so a fresh worktree needs `--features vcad-kernel-text/no-builtin-font` (what CI uses) or an `npm ci`.


ecto:
Rebased onto main and extended — the merge changed what this PR needs to cover.

**`photoreal.rs` exists after all.** My branch was 5 commits behind main, which had since landed the photoreal path tracer (`crates/vcad-render/src/photoreal.rs` + `crates/vcad-kernel-raytrace/src/pathtrace.rs`). My earlier note in the PR description that the file didn't exist was true of the stale base, not of main. Its `rasterize` carries the identical `as_brep()`-or-skip, so it's now wired the same way: mesh solids build a crease-baked triangle BVH, and untraceable parts are named in an error rather than dropped.

The photoreal coverage test reads the **alpha channel** rather than comparing to a background colour — the path tracer lights the scene, so "not the background colour" isn't a reliable coverage signal, but a shadow-catcher backdrop leaves un-hit pixels transparent.

**On the red `Rust (nightly)` check:** that failure is pre-existing on main, not from this PR. It's `clippy::too_many_arguments` on `pathtrace.rs`'s `sample_lights`, and it fails main's own `Rust (stable)` and `Rust (nightly)` jobs on the current tip. It arrived here via the merge. I've silenced it with a scoped `#[allow]` and a note on why — the shading frame is integrator hot-loop state, so bundling it into a struct to satisfy the lint would add a copy per light sample. Happy to drop that hunk if you'd rather fix it separately on main.

Local verification on the merged tree:

```
cargo test -p vcad-kernel-raytrace -p vcad-render     # 74 + 74 lib, + integration, all green
cargo clippy -p vcad-kernel-raytrace -p vcad-render --all-targets -- -D warnings   # clean
cargo fmt --check   # clean
```

PR #745, opened 2026-07-29
feat(mcp): durable Tasks-extension records across server restarts

ecto:
## What

Stacked on #743 (retarget to `main` after it merges). Completes the first follow-up from that PR: **Tasks-extension records now survive server restarts** — without pretending a half-run background tool can resume (it can't; the in-process bridge dies with the process).

Three guarantees:

1. **Terminal tasks survive.** When a task reaches completed/failed/cancelled, the serializable record (taskId, status, statusMessage, timestamps, toolName, result, error) persists to a durable task store (`packages/mcp/src/task-store.ts`, mirroring the session-store pattern exactly): file store under `<sessionDir>/tasks` locally, Supabase `mcp_tasks` (migration 037, capability-keyed + service-role-only like `mcp_artifacts`) when the env is present, in-memory no-op otherwise. `tasks/get` hydrates from the store on a cache miss before reporting unknown, and re-caches.
2. **Orphaned in-flight tasks report honestly.** taskIds now embed the minting process's boot token (`task_<tok>_<random>`, same scheme as session ids). A prior-boot taskId with no durable terminal record gets a `status:"failed"` task result explaining the server restarted mid-run and the call should be re-issued — not an unknown-id error. A same-boot unknown id still errors as a typo.
3. **Durability honesty.** When `isSessionStoreDurable()` is false, the `CreateTaskResult` statusMessage notes the handle will not survive a restart (the session-minting tools' pattern).

## Not included

- Migration 037 is **not pushed to prod** — deploy with `supabase db push`.
- The other #743 follow-up (real `notifications/progress` from router/solvers) needs progress hooks through the kernel/WASM boundary and stays deferred.

## Verification

- 7 new tests in `protocol-2026.test.ts` (env-isolated temp dir, per the durability-smoke pattern): terminal round-trip through a simulated restart, failed-by-restart orphan reporting on get and cancel, typo-vs-restart distinction, TTL expiry through the store, and both durability-note branches
- Full suite: 1081 passed / 97 files; `tsc --noEmit` clean; `deprecated-surface` tripwire green

### DIFF
--- a/changelog/entries/2026-07-29-durable-task-records.json
+++ b/changelog/entries/2026-07-29-durable-task-records.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-29-durable-task-records",
+  "version": "0.9.4",
+  "date": "2026-07-29",
+  "category": "fix",
+  "title": "Task handles survive server restarts",
+  "summary": "Finished MCP task records persist across restarts, and a taskId orphaned mid-run reports the restart honestly instead of 'Unknown taskId'.",
+  "features": ["mcp", "tasks", "durability"],
+  "mcpTools": []
+}
--- a/packages/mcp/src/__tests__/protocol-2026.test.ts
+++ b/packages/mcp/src/__tests__/protocol-2026.test.ts
@@ -23,6 +23,7 @@ import {
   isModernMessage,
   handleModernRequest,
   taskIdBootToken,
+  flushTaskPersists,
 } from "../protocol-2026.js";
 
 /**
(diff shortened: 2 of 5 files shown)


ecto:
Addressed the Choji findings in e8446dde: terminal-record persists are now tracked in a module-level set with an exported `flushTaskPersists()`, and the restart round-trip test awaits it instead of a 100 ms sleep — no timing dependence, and shutdown/test callers can drain pending writes. (An in-memory prune racing a persist was already safe: prune only drops the warm cache; the store write completes independently, and the file store's write-then-rename can't leave a torn record.)

PR #747, opened 2026-07-29
feat(mcp): stream real progress from nine long-running kernel tools

ecto:
Stacked on #746 — base that branch, so this diff is the single commit `edc83381`. Merge #746 first.

## The problem

#746 gave `notifications/progress` to every tool whose iteration loop lives in TypeScript. The rest were each **one synchronous kernel WASM call** whose loop runs in Rust — `fab_prep`, `topology_optimize`, and the `simulate_*` family. They go silent for minutes, then report everything at once.

A `js_sys::Function` callback invoked from inside that synchronous call **cannot** fix this. The Node event loop is blocked for the whole call, so every queued notification bursts out only after completion — fake streaming.

## The fix

Chunked kernel APIs: an env/handle created once, plus a `step`/`advance` method the TS layer calls in a loop, emitting `ctx.progress` between calls while the event loop is actually free. This is the seam `MdEnv.run` already proved out.

Each tool gets the same four layers — a stepwise driver in its own kernel crate, a wasm run class, an `@vcad/engine` wrapper that **falls back to the one-shot call on an older kernel**, and `ctx.progress` wiring in the handler.

| Tool | Progress granularity |
|---|---|
| `fab_prep` | per strip-and-re-route round |
| `topology_optimize` | per SIMP iteration (carries compliance + density change) |
| `simulate_flow` | per chunk of LBM timesteps (carries steadiness residual) |
| `simulate_photonics` | per chunk of FDTD timesteps |
| `simulate_charged_particles` | per chunk of traced particles |
| `optimize_electrodes` | per multi-start FD evaluation |
| `simulate_neutron_shield` | per batch of histories |
| `simulate_lattice_gauge` | per chunk of MC sweeps |
| `minimize_energy` | per chunk of FIRE iterations |

## Why the numbers can't drift

Chunked results are bit-identical to the one-shot path **by construction**: wherever possible the one-shot entry point is reimplemented on top of the stepwise driver (`run_fab_prep`, `solve_steady`, `optimize`), so there is no second code path to drift out of sync. RNG state lives in the env, so the Monte Carlo tools chunk without perturbing their stream. `FlowStepper`'s steadiness checks key off the **absolute** step count, so chunk boundaries can't change when it converges.

Every crate carries a parity test asserting one-shot == chunked on the same input:

```
vcad-ecad-fabprep     stepwise_session_matches_the_one_shot_run
vcad-kernel-topopt    stepwise_run_matches_the_one_shot_optimize
vcad-kernel-flow      chunked_stepper_matches_the_one_shot_solve
vcad-kernel-photonics chunked_run_matches_one_shot_run          (bit-level, .to_bits())
vcad-kernel-particle  chunked_ensemble_ranges_match_the_one_shot
                      stepwise_multi_start_matches_the_one_shot
vcad-kernel-neutronics chunked_run_matches_the_one_shot
vcad-kernel-qcd       chunked_run_matches_the_one_shot
vcad-kernel-atoms     stepwise_fire_matches_phyz_fire
```

## Reviewer notes

- **`vcad-kernel-flow/src/solve.rs` is the big diff** and the one worth real attention. `solve_steady`'s ~500-line loop became `FlowStepper::{new, advance, into_solution}`; the body is unchanged apart from field access through `&mut self`. `solve_steady` is now `advance(usize::MAX)`.
- The engine wrappers degrade gracefully: a kernel WASM predating a run class silently takes the one-shot path, so this does not hard-require a kernel refresh.
- No `packages/kernel-wasm/vcad_kernel_wasm*` artifacts touched — `wasm-refresh.yml` still owns them.

## Not included

`simulate_em` and `homogenize_material` still take the one-shot path. Same pattern applies; left for a follow-up rather than shipped half-wired.

## Verification

- `cargo clippy -D warnings` clean across all nine touched crates + `vcad-kernel-wasm` (fixed 3 pre-existing `needless_range_loop` in neutronics)
- `cargo test` green on all touched crates, including `vcad-kernel-flow` (35 passed, 347s)
- `packages/mcp`: `tsc --noEmit` clean, `vitest run` 1078 passed / 3 skipped across 97 files
- `cargo fmt` applied per-crate

### DIFF
--- a/changelog/entries/2026-07-29-kernel-chunked-progress.json
+++ b/changelog/entries/2026-07-29-kernel-chunked-progress.json
@@ -1,0 +1,22 @@
+{
+  "id": "2026-07-29-kernel-chunked-progress",
+  "version": "0.9.4",
+  "date": "2026-07-29",
+  "category": "feat",
+  "title": "Live progress from long-running kernel tools",
+  "summary": "Nine kernel-bound tools now stream real notifications/progress while they run, instead of going silent for minutes and reporting everything at the end.",
+  "features": [
+    "mcp"
+  ],
+  "mcpTools": [
+    "fab_prep",
+    "topology_optimize",
+    "simulate_flow",
+    "simulate_photonics",
+    "simulate_charged_particles",
+    "optimize_electrodes",
+    "simulate_neutron_shield",
+    "simulate_lattice_gauge",
+    "minimize_energy"
+  ]
+}
--- a/crates/vcad-ecad-fabprep/src/lib.rs
+++ b/crates/vcad-ecad-fabprep/src/lib.rs
@@ -301,144 +301,224 @@ fn refused(pcb: &Pcb, opts: &FabPrepOptions, why: String) -> FabPrepOutcome {
     }
 }
 
-/// Run the whole fab-prep pipeline against `pcb`, mutating it in place.
-///
-/// Never panics on a difficult board: a run that cannot converge returns a
-/// report with `converged: false` and the offenders named, and it is the
-/// caller's job (see [`package::write_fab_package`]) not to ship it.
-pub fn run_fab_prep(pcb: &mut Pcb, opts: &FabPrepOptions) -> FabPrepOutcome {
-    // 0. Waivers. A name that matches no rule is refused outright: a typo in a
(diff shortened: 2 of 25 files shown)


ecto:
@chojiai ping

PR #749, opened 2026-07-30
fix(physics): real part geometry, correct multi-DOF integration, gravity-comp servos

ecto:
## What

Fixes the non-physical gym dynamics reported for revolute joints (anti-restoring gravity, sibling "torque leak", stale `end_effector_poses`) — plus two deeper defects found while verifying.

1. **Composed parts no longer collapse to a placeholder cube.** `PhysicsWorld::evaluate_part` matched primitive `CsgOp`s only and fell through to `Solid::cube(10,10,10)` for anything composed (`union`, `translate`, …). The placeholder's COM sat *above* the joint anchor, so every pendulum was secretly inverted — that one substitution produced both reported symptoms. Parts now evaluate through `vcad_eval` (cached per PartDef root). Unresolvable geometry fails closed unless the PartDef carries authored inertials (the URDF `package://` case), which get a *centred* placeholder collider.

2. **Integration is now joint-aware.** `step()` hand-rolled `q += v*dt`, which phyz's own integrator docs call out as wrong for multi-DOF joints: free-floating bodies **never fell** (gravity accumulated in the linear `v` slots while position integrated the angular slots), and ball-joint exp-coords were added instead of composed. Swapped for `phyz::rigid::semi_implicit_euler`. `step()` also dropped its FK result, which is why end-effector poses never moved.

3. **Passive rollouts are actually passive; servos actually reach their target.** `from_document` installed a PD servo on any joint authored at a nonzero `state`; it now seeds `q` only. Position/velocity motors gain one RNEA gravity-feedforward pass per substep (torque mode untouched — RL actions stay raw): a 70° command now settles at 70.000° instead of drooping to 42.8° (`τ_g/kp`).

## Verification

New oracle suites `tests/revolute_dynamics.rs` and `tests/integrator_probes.rs`, all pinned to closed form:

- compound-pendulum period vs `T = 2π√(I_o/mgd)`: **0.017% error** (rules out the suspected mm-vs-m inertia bug — there is none)
- ballistic free fall, exact spin conservation (quat = (cos 4, 0, 0, sin 4) after 1 s @ 8 rad/s)
- ball-pendulum climb exactly matches its kick's energy budget
- 60 s passive rollout amplitude pinned at 45.005° (symplectic, no drift)
- FK arc centred on the joint anchor through 0/30/90° (rules out the world-origin-anchor FK bug for the physics path)

`cargo test --workspace --exclude vcad-desktop` green; clippy/fmt clean.

## Reviewer notes

- **Merge ordering: requires the companion phyz fix** (spherical/free `joint_transform_slice` returned the rotation instead of its inverse coordinate map, on CPU and both GPU shaders — the energy pump behind phyz's "half-cheetah contact instability"). That change is in the local phyz checkout, uncommitted; CI clones phyz at main tip, so `integrator_probes::ball_joint_pendulum_is_bounded_and_restoring` will fail here until it lands in ecto/phyz.
- MCP users won't see this until `wasm-refresh.yml` rebuilds the kernel WASM on main (no artifacts committed here, per the single-writer rule).
- `set_joint_velocity_raw` / `set_free_body_spin_raw` are `#[doc(hidden)]` test hooks.


ecto:
Reopened. The earlier red was a **timing artifact, not a real failure**.

The only failing test was `ball_joint_pendulum_is_bounded_and_restoring` ("ball pendulum climbed above the anchor") — the ball-joint energy pump fixed upstream by ecto/phyz#34. That PR merged at `20:31:41Z`; this run started at `20:16Z`, so CI cloned phyz **before** the fix landed and the probe correctly caught the old bug. This PR was closed at `20:31:43Z`, two seconds after the upstream merge.

Now merged with `origin/main` (was 13 behind) and re-verified locally against merged phyz:

- `cargo test -p vcad-kernel-physics` — all green, including `integrator_probes` 6/6 with `ball_joint_pendulum_is_bounded_and_restoring` passing
- `cargo clippy -p vcad-kernel-physics --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean
- no `packages/kernel-wasm/vcad_kernel_wasm*` changes, so `wasm-artifact-guard` stays green

Verified independently against prod MCP (`0.9.4+85a6987`) that the three reported symptoms still reproduce byte-for-byte on `main` — passive `ja` 60→+80 / `jb` 0→−80, and the `1.693627°` / `339.7117 °/s` sibling leak — confirming this is still the live fix for them.

Note for after merge: `create_robot_env` hands raw IR to `PhysicsWorld`, so MCP users won't see any of this until the kernel WASM is refreshed on main.

ecto:
Resolved the merge conflict with `main` (#759 "Add ground-plane contact" landed in the same `world.rs` `step()`).

**Both sides had independently fixed the same two bugs**, so main's block was a strict superset — it already used the joint-aware `phyz::rigid::semi_implicit_euler` instead of a hand-rolled `q += v·dt`, and already published the FK result into `state.body_xform`. Took main's ground-contact block wholesale and verified it's semantically identical to mine when ground is disabled: mine ran `semi_implicit_euler(qdd)` (updates `v`, then integrates `q`); main computes `free_v = v + qdd·dt`, assigns it, then integrates positions with `zero_qdd` — same result.

Kept from this branch:
- the comment explaining *why* the FK publish matters (main's side was bare)
- the ball-joint clause in the integrator rationale — main's comment only covered Free joints, and ball joints are exactly what `ball_joint_pendulum_is_bounded_and_restoring` exercises

Everything else on this branch (`evaluate_part` through `vcad_eval`, `joints.rs`, both new test files) auto-merged untouched. Net diff vs `main` is 6 files — only this fix, none of main's work pulled back in.

Local verification on the merge commit:
- `cargo test -p vcad-kernel-physics` — all green, including main's new `unitree_g1_pd_hold_with_ground_contact_stays_finite`, plus `integrator_probes` 6/6 and `revolute_dynamics` 5/5
- `cargo clippy -p vcad-kernel-physics --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean
- no `packages/kernel-wasm/vcad_kernel_wasm*` changes

PR #751, opened 2026-07-30
Un-mirror every drafting view — shop drawings showed the opposite side of the part

ecto:
## What

`crates/vcad-kernel-drafting` rendered every principal view (Front, Back, Top, Bottom, Right, Left) and the isometric view as a **mirror image of the opposite side of the part**. Dimensioned shop drawings produced from these views were unusable for manufacturing.

This is the drafting-crate sibling of the isometric mirror fixed in `vcad-render` by #728 — that PR's commit message explicitly called this defect out as "deserves its own change."

## Root cause — worse than the doc comments admitted

`types.rs` documented Back/Bottom/Left as "mirrored," implying three bad views. In fact **all six were wrong**, via a sign clash between two files:

- `ViewDirection::view_vector()` returned a vector pointing **viewer → scene**.
- `ViewMatrix::from_view_direction`'s basis math (`right = up × forward`, `up = forward × right`) and `Triangle3D::is_front_facing` both assume it points **scene → viewer**.

The two errors composed into a self-consistent camera sitting on the *opposite* side of the part. So `ViewDirection::Front` drew a mirrored *back* view — with hidden-line visibility computed for the wrong side too, meaning the wrong edges were dashed. Self-consistent across a sheet, which is exactly why it survived: nothing looked broken, it was just wrong.

## Changes

**`types.rs`** — `view_vector()` now genuinely points scene → viewer. Front/Back/Top/Bottom flipped; isometric elevation sign fixed so the camera is above the part, matching `vcad-render`'s `(1,1,1)` convention. Notably **Right/Left were already correct** under the fixed reading — their labels only *appeared* swapped when seen through the mirror, so flipping them would have been the wrong fix. Doc comments now state the actual world→paper axis mapping per view instead of the word "mirrored."

**`projection.rs`** — the cross-product basis formulas are correct for a viewer-ward `forward` and are **unchanged**; the code now matches its own doc comments. Docs clarify that depth increases toward the viewer.

**`hidden_line.rs`** — depth now grows toward the viewer, so the two occlusion comparisons (the broadphase reject and the interpolated-depth test) flipped accordingly.

**`SectionPlane::right`** — had a `-X` normal, i.e. a left-side frame mislabeled as right; now `+X`, consistent with `ViewDirection::Right`. `front`/`horizontal` were already correct.

## Tests

Following #728's handedness-pinning pattern:

- `handedness_is_pinned_for_every_view` — `right × up · forward == +1` for all 6 principal views, standard iso, dimetric, and both degenerate straight-down/straight-up fallback paths. The drafting analogue of render's `view_basis_handedness_is_pinned`.
- `principal_view_axes_are_pinned` — locks each of the 6 views to exact third-angle convention (Front: +X right / +Z up; Right: +Y right; Bottom: −Y up; etc.). Pins the *specific* axes, not merely "some orthonormal basis."
- **`tests/view_orientation.rs`** (new) — the regression the issue asked for:
  - An **L-bracket with an off-center hole** (hole far out on the +X leg): the hole must land on the correct side of the drawing in each of the 6 views.
  - A **block in front of a slab**: must show visible edges in the Front view and be fully occluded in the Back view. This catches the visibility half of the bug, which a pure-geometry test would miss.

  Both fail on the old basis.

The three sign-ambiguous assertions in `projection.rs` that hid the bug (`assert!((p2.x.abs() - 1.0).abs() < 1e-10)` — magnitude only, "may be negated depending on coordinate system") are now exact signed values.

## Goldens

Only one golden existed: `tests/golden/shop_drawing.pdf`. Its subject is a **symmetric 60×40×20 block**, so the byte delta is purely the coordinate-frame sign change and no mirrored geometry was baked in — verified before regenerating with `REGEN_GOLDEN=1` rather than updating blind.

## Downstream

Checked `hidden_line.rs`, `section.rs`, `sheet.rs`, `dimension/`, `pdf.rs`, `detail.rs`, `edge_extract.rs`, the WASM `project_view` / `project_mesh_wasm` bindings, `vcad`'s DXF export, `vcad-kernel-step`'s PMI, and the app's `drawing-store` / `DrawingView`. All consume `ProjectedView` as pass-through with **no compensating flips** to unwind, so they inherit the fix unchanged.

## Verification

- `cargo test -p vcad-kernel-drafting` — 111 tests green
- `cargo test -p vcad-kernel-step` — green
- `cargo fmt -p vcad-kernel-drafting`

Note: `cargo test -p vcad` cannot build in this worktree due to the pre-existing `vcad-kernel-text` `include_bytes!` path into `node_modules` (unrelated to this change; a worktree-depth issue).

Changelog entry added at `changelog/entries/2026-07-30-drafting-views-unmirrored.json` (category: `fix`).


ecto:
@chojiai Addressed the back-view occlusion nit in `be2e91e0`.

You were right that the match was under-constrained. It required *both* endpoints inside the square, so an edge straddling the boundary could slip past. It now matches on **either** endpoint (a strictly wider net) and additionally asserts the match set is **non-empty** before requiring every edge in it to be hidden — that second part closes a failure mode neither of us named: the old check would have passed vacuously if the world→paper axis mapping regressed and the block projected somewhere else entirely.

Re-verified the hardened test is still red on the pre-fix mirrored basis, so widening the net didn't make it vacuous.

ecto:
@chojiai Thanks for the re-read. On the line ~149 nit — no change needed, that one's an artifact of the diff excerpt rather than the file.

The comment is complete at `be2e91e0`; it's just wrapped across two lines:

```rust
// Guards against the check silently matching nothing (e.g. if the
// world → paper axis mapping regressed and the block moved elsewhere).
```

Your quote cut at `(e.g. if th`, which is the first line's end. Verified against `git show be2e91e0:crates/vcad-kernel-drafting/tests/view_orientation.rs` — and a genuinely truncated comment would have been a syntax-clean but nonsensical sentence, not a build failure, so worth confirming directly rather than inferring from CI. Rust stable + nightly both went green on that SHA.

Leaving the code as-is.

PR #753, opened 2026-07-30
fix(sketch): report pending, not solved, when the kernel isn't loaded

ecto:
## What

`solveSketch` in `packages/core/src/stores/sketch-store.ts` had a fallback for when `getKernelWasmSync()` returns `null` (or the module lacks `solveSketchSegments`) — i.e. the kernel WASM hasn't hydrated yet. That fallback set `solved: true` and `constraintStatus: "solved"` **without running the solver or touching any segments**. The badge told the user their sketch was fully constrained when in fact nothing had been solved.

This replaces the dishonest state with an honest one:

- New `"pending"` variant on `ConstraintStatus` (`packages/core/src/types.ts`).
- The fallback now sets `solved: false, constraintStatus: "pending"` and kicks off `getKernelWasm()`. When the kernel hydrates, it re-runs `solveSketch()` if the sketch is still active and still pending — so a solve requested during app boot completes for real on its own rather than being silently dropped. If the kernel never loads (headless, tests), the `catch` leaves it honestly pending.
- The two components that color the status badge — `FeatureTree.tsx` and `StatusBar.tsx` — render `pending` as dimmed neutral gray, clearly distinct from the emerald "solved" state.

## Why

A constraint status is a correctness claim about the user's geometry. Reporting "solved" for a solve that never ran is the kind of thing a user only discovers downstream, when the extrude comes out wrong.

## Notes for reviewers

- No existing tests depended on the old fallback path — the only `solveSketch` callers are `useAppCommands.ts` and `useToolDefinitions.tsx` in the app, neither of which asserts on status.
- The re-solve guard checks both `s.active` and `s.constraintStatus === "pending"` so a user who has since exited the sketch, or whose status has moved on, doesn't get a surprise geometry rewrite after hydration.
- `npm run build -w @vcad/core -w @vcad/app` passes; `npm test --workspaces --if-present` is green across all workspaces.
- Changelog entry added (`fix`) since the visible badge behavior changed.

### DIFF
--- a/changelog/entries/2026-07-30-sketch-solve-pending-status.json
+++ b/changelog/entries/2026-07-30-sketch-solve-pending-status.json
@@ -1,0 +1,9 @@
+{
+  "id": "2026-07-30-sketch-solve-pending-status",
+  "version": "0.9.4",
+  "date": "2026-07-30",
+  "category": "fix",
+  "title": "Honest sketch solve status before kernel loads",
+  "summary": "Solving a sketch before the kernel WASM finishes loading now shows a dimmed 'pending' badge and re-solves automatically once it loads, instead of falsely showing 'solved'.",
+  "features": ["sketch", "constraints"]
+}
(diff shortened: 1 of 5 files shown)


ecto:
Addressed the minor finding in `5c8565bd`: the post-hydration re-solve is now wrapped in a try/catch that logs and falls to `"error"`, so a throw from the second solve no longer leaves the sketch parked in `"pending"` with nothing visible. A kernel that genuinely never hydrates still stays honestly pending, since in that case nothing was solved and nothing is coming.

PR #754, opened 2026-07-31
fix(kernel): booleans on filleted solids trim along the true arc

ecto:
Closes both fillet defects that were parked behind `#[ignore]`d tests.

## The planar-face trim defect

The planar side of a fillet seam only conformed to the cutting circle when that circle crossed the face boundary **exactly twice**. Everything else fell through, leaving the face chorded while the curved partner followed the arc — the shell zippered open and the volume came out ~21% high. It turned out to be three cooperating cases in `split.rs`, not one:

1. **>2 crossings** — `split_planar_face_by_multi_arc` traces alternating boundary/arc regions (Weiler–Atherton style), emitting one sub-face per polygon∩disk and polygon∖disk region. Fails back to the unsplit face *before any mutation* on a non-alternating pattern.
2. **Inscribed circle** (4 tangencies, 0 crossings) — the dominant case here, and one the original bug report didn't name. The cutter's planar faces are first gridded by the edge-cylinders' `TwoLines` splits, leaving a square cell in which each corner-sphere circle is inscribed, so nothing could split it. `split_planar_face_inscribed_circle` emits the disk plus one sliver per tangent gap; `insert_circle_tangent_vertices` keeps the neighboring cell from T-junctioning at the tangent point.
3. **Spherical patches with >2 crossings** — `clip_spherical_face_by_circle` declined and fell through to the *destructive* whole-sphere split, which replaced an already-trimmed patch with two full caps (resurfacing deleted geometry as an unmatched full-circle rim). `clip_spherical_face_multi` traces the same regions on the sphere and returns the face **unchanged** rather than `None` on an untraceable pattern, since `None` is what triggers the destructive path.

**Seam-schedule gotcha worth knowing:** sphere-sourced circles ride the raw `θ = 2πk/segments` grid in the circle's own frame, *not* `canonical_arc_points`' sag-adaptive schedule (the cylinder convention). The new planar arcs sample through a shared `circle_frame_arc_points` to match, or they zipper against the sphere clips.

`difference_with_filleted_subject_is_watertight` now passes: **0 open edges** and **86,712 mm³** against the 86,618 Monte-Carlo truth (0.11%), from 568 open edges and ~105,000 mm³.

## The adjacent-edge corner blend

This one was stale rather than broken. `fillet_subset`'s miter-chain path already handles vertex-sharing edges: at a trihedral mirror-symmetric corner the two equal-radius blend cylinders intersect exactly in a planar curve on the bisector plane, so each is trimmed there and both ends share one sampled curve — watertight by construction, no torus/sphere patch needed. Only the `#[ignore]` and the doc comment needed updating. The test passed unmodified.

## Winding pin

`Solid::volume()` takes the absolute value of the signed divergence-theorem sum, so a globally inverted mesh — the reported K1 humanoid symptom, filleted parts turning inside-out under difference — satisfies **every** existing volume assertion silently. `boolean_against_filleted_subject_keeps_outward_winding` computes the signed sum directly on both a difference and a union.

## Verification

- `cargo test -p vcad-kernel-fillet -p vcad-kernel` — all green. The only remaining `#[ignore]`s in `vcad-kernel` are the two intentional porkchop diagnostics (`diag_porkchop_boundary_edges`, `diag_porkchop_fillet_trace`), which are print-only tools, not defects.
- `cargo clippy --workspace --exclude vcad-desktop -- -D warnings` — clean.
- `cargo fmt` applied to touched files.
- No `packages/kernel-wasm/vcad_kernel_wasm*` artifacts touched.

## For the reviewer

`split.rs` is +936 lines and is where the risk lives. The region tracers are the part worth the closest read — both are written to degrade to a no-op rather than a partial mutation, but the fallback ordering (bail before mutating; return `Some(unchanged)` never `None` on the sphere side) is load-bearing and easy to break later.

### DIFF
--- a/changelog/entries/2026-07-30-fillet-boolean-conformity.json
+++ b/changelog/entries/2026-07-30-fillet-boolean-conformity.json
@@ -1,0 +1,13 @@
+{
+  "id": "2026-07-30-fillet-boolean-conformity",
+  "version": "0.9.4",
+  "date": "2026-07-30",
+  "category": "fix",
+  "title": "Booleans on filleted solids are watertight",
+  "summary": "Cuts that slice fillet blends now trim planar faces along the true arc (multi-crossing, inscribed, and tangent cases), so differences against filleted parts come out watertight with correct volume and outward windings.",
+  "features": [
+    "fillet",
+    "booleans"
+  ],
+  "mcpTools": []
+}
--- a/crates/vcad-kernel-booleans/src/split.rs
+++ b/crates/vcad-kernel-booleans/src/split.rs
@@ -800,7 +800,12 @@ pub fn split_planar_face_by_circle(
         if circle_partially_inside_polygon(&loop_verts, circle) {
             return split_planar_face_by_arc(brep, face_id, circle, segments);
         }
-        // Circle doesn't intersect at all
+        // The circle doesn't cross this face, but it may be TANGENT to one
+        // of its edges from outside (the neighbor of a cell the circle is
+        // inscribed in). The inscribed split puts a vertex at every tangent
+        // point, so this face's straight edge must gain the same vertex or
+        // the shared boundary T-junctions and the shell zippers open.
+        insert_circle_tangent_vertices(brep, face_id, circle);
         return SplitResult {
             sub_faces: vec![face_id],
         };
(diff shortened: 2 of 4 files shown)


ecto:
@chojiai both findings addressed in 7da04269, along with a torture-track regression that surfaced separately.

- **Major — inscribed-circle path unreachable below 4 contacts:** correct, and the parity check rejected `m = 3` too, so a triangular cell with an inscribed circle was doubly blocked. The floor is now the real minimum of 3, and the tracer's own preconditions (`m >= 4`, even) moved down to after the contacts are classified — so an odd all-tangency count reaches the inscribed path instead of being rejected for the crossing tracer's reasons.
- **Minor — zero-length span in `circle_frame_arc_points`:** now an explicit early return with a comment. Your read was right that it was harmless (the caller's `len < 3` bail caught it), so this is clarity, not a behavior change.

Separately, the torture track caught a real regression this PR introduced — `sliv-cyl-graze-t1-d`, pass → 5 open boundary edges. A cylinder grazing a cube face by 1e-6 falls inside `insert_circle_tangent_vertices`' 1e-6 tolerance, and that function was willing to *mint* a tangent vertex rather than only adopt one the inscribed split had already created. On a graze nothing is split across the edge, so it put a vertex on one side and left the other alone — the exact T-junction the function exists to prevent. It now skips any tangency with no existing vertex at that point. Track is back to 0 regressions (297 → 298 pass).

PR #755, opened 2026-07-31
fix(step): STEP import reports skipped faces instead of silently dropping them

ecto:
## What

A STEP face whose surface fails `parse_surface_oriented` with `UnsupportedEntity` was added to `skipped_faces` in `read_solid` and omitted from the imported solid **with no signal at all** — the user got a solid with holes in it and no way to find out why, or which faces were lost.

Import now threads a report out of the reader and surfaces it at every consumer, matching the repo-wide fail-closed reporting convention (see `vcad-receipt`).

## The report

`SolidImportReport` carries, per solid: the shell's `total_faces`, and one `SkippedFace { face_id, surface_id, reason }` per omission — the STEP entity id of the face, the entity id of the surface it referenced, and the surface type name lifted from the parse error. `StepImportReport` wraps the per-solid records and offers `is_clean()`, `total_skipped_faces()`, and a ready-to-print `summary()`.

Chordal subdivision of curved edges lands in a per-solid `notes` list rather than per face — it approximates geometry but doesn't remove any, so it doesn't warrant a face-level entry.

## API

`read_step_with_report` / `read_step_from_buffer_with_report` return `ReadResult { solids, report }`. The existing `read_step` / `read_step_from_buffer` become thin wrappers over them, so **every existing caller keeps compiling unchanged**. `vcad-kernel` mirrors the pair with `Solid::from_step_all_with_report` / `from_step_buffer_all_with_report`.

## Consumers

- **`vcad-cli import-step`** — prints a stderr warning summary naming each skipped face and its surface type.
- **`import_step` (MCP)** — adds `summary.warning` plus a structured `summary.skipped_faces` array to the result text when anything was dropped.
- **App drag-drop** — raises a warning toast (`STEP import skipped N faces with unsupported surfaces — the geometry has holes there`) and logs the full per-face detail.
- **WASM** — new `importStepBufferWithReport` returning `{ meshes, report, summary }`.

## Reviewer notes

- **No `packages/kernel-wasm/vcad_kernel_wasm*` artifacts are committed** — `wasm-refresh.yml` on main owns those. Because of that, the checked-in WASM predates the new binding, so `Engine.importStepWithReport` feature-detects `importStepBufferWithReport` and falls back to the bare import with an empty report. The app path degrades gracefully and lights up once CI refreshes the artifacts after merge.
- The two `artifact-store.test.ts` import_step fixtures were updated because their engine mocks stubbed `importStep`, which the tool no longer calls.

## Testing

- New `crates/vcad-kernel-step/tests/import_report.rs`: a hand-written STEP fixture whose lateral face uses `DEGENERATE_TOROIDAL_SURFACE` (a real AP214 entity the reader doesn't support) asserts the report names the face (`#29`), the surface (`#40`), and the type; plus a clean-import case and a compat-wrapper case.
- New MCP test asserting the report reaches the tool result.
- `cargo test --workspace` — exit 0.
- `cargo clippy --workspace --exclude vcad-desktop -- -D warnings` — clean.
- `npm test -w @vcad/mcp` — 97/97 files, 1079 passed.
- Builds green for `@vcad/engine`, `@vcad/mcp`, `@vcad/app`.

### DIFF
--- a/changelog/entries/2026-07-30-step-import-skipped-face-report.json
+++ b/changelog/entries/2026-07-30-step-import-skipped-face-report.json
@@ -1,0 +1,14 @@
+{
+  "id": "2026-07-30-step-import-skipped-face-report",
+  "version": "0.9.4",
+  "date": "2026-07-30",
+  "category": "fix",
+  "title": "STEP import reports skipped faces",
+  "summary": "STEP import now reports faces skipped for unsupported surface types (entity id + surface type) instead of silently dropping them — surfaced in the CLI, the app's import toast, and the import_step MCP tool.",
+  "features": [
+    "step-import"
+  ],
+  "mcpTools": [
+    "import_step"
+  ]
+}
--- a/crates/vcad-cli/src/main.rs
+++ b/crates/vcad-cli/src/main.rs
@@ -810,7 +810,15 @@ fn import_step(input: &PathBuf, output: &PathBuf, name: Option<String>) -> Resul
     });
 
     // Import the STEP file
-    let solids = Solid::from_step_all(input)?;
+    let (solids, report) = Solid::from_step_all_with_report(input)?;
+
+    if let Some(summary) = report.summary() {
+        eprintln!(
+            "warning: STEP import skipped {} face(s) with unsupported surface types:",
+            report.total_skipped_faces()
+        );
+        eprintln!("{summary}");
+    }
 
     if solids.is_empty() {
         anyhow::bail!("No solids found in STEP file");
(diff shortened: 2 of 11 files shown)


ecto:
Thanks — both observations were accurate. One fixed, one left alone with reasoning:

**Toast count derived by regex** — fixed in 9dab451b. The count now sums `skipped_faces` across the structured `report` instead of matching `/face #/g` against the prose summary, and the toast is gated on that count rather than on `summary` being non-null. The summary string is still what gets logged to the console, but it no longer drives any logic, so rewording `StepImportReport::summary()` can't silently change the number shown.

**`remote.test.ts` `fakeEngine` is `{}`** — left as-is. This predates the PR: the mock was already `{}` on `main`, where the same three tests called `importStep`, which would have thrown `engine.importStep is not a function` in exactly the same way had they reached it. All three exercise argument-validation paths (hosted-mode path rejection, neither-arg, empty base64) that throw before any engine call, so the empty mock is deliberate rather than incidental, and swapping `importStep` for `importStepWithReport` neither improved nor worsened it. Giving it a real stub seemed more likely to imply those tests touch the engine than to prevent a future mistake.

App build and the full MCP suite (97 files / 1079 tests) pass on the updated commit.

PR #760, opened 2026-07-31
Sim2real gym: domain randomization and richer step results

ecto:
## Summary

Extends the robot gym environment with seeded domain randomization, observation noise, configurable termination conditions, and richer per-step diagnostics. Enables reproducible sim2real training workflows where identical seeds produce identical randomized episodes.

## Key Changes

**Core gym API:**
- Added `EnvConfig` struct for domain randomization, observation noise, termination conditions, and base instance selection
- New `RobotEnv::new_with_config()` constructor accepting explicit config; `new()` delegates to it with defaults
- New `RobotEnv::reset_with_seed()` to re-seed the domain-randomization stream and reset in one call
- `step()` now returns `StepResult` with full diagnostics via `step_full()`; legacy `step()` signature preserved for compatibility

**Domain randomization (seeded, per-episode):**
- Per-link multiplicative mass scaling (models density variation)
- Per-joint friction and viscous damping scaling
- Global PD motor gain scaling
- Actuator latency as integer physics substeps (delay line queues actions)
- Initial joint position and velocity perturbations

**Observation enhancements:**
- Added `base_pose` and `base_velocity` to `Observation` (configurable base instance, defaults to ground)
- Gaussian observation noise on joint positions/velocities and base state (std-dev per component)
- Noise applied only to `step`/`reset` observations; `observe()` stays noise-free

**Termination conditions:**
- Configurable base height threshold, base tilt angle, and joint-limit triggers
- Legacy behavior (end-effector below -1m) preserved when no `TerminationConfig` is set
- Per-step diagnostics include termination reason, base height/tilt, and joint limit violations

**Physics world updates:**
- `PhysicsWorld::refresh_kinematics()` to recompute FK and spatial velocities after direct state mutations
- `scale_instance_mass()`, `scale_joint_friction()`, `set_gain_scale()`, `perturb_joint_state()` for domain randomization
- Cached `body_vels` and `gain_scale` fields

**WASM bindings:**
- `PhysicsSim::new()` accepts optional `config_json` parameter
- New `resetSeeded(seed)` method for seeded resets
- `step()` returns `StepResult` with `info` field

**TypeScript/MCP:**
- New `PhysicsEnvConfig`, `PhysicsDomainRandomization`, `ObservationNoise`, `TerminationConfig` interfaces
- `createRobotEnv()` accepts optional `config` parameter
- `gymReset()` accepts optional `seed` parameter
- `PhysicsStepResult` now includes optional `info: PhysicsStepInfo`
- Comprehensive test suite (`gym-sim2real.test.ts`) validating reproducibility, latency, and diagnostics

## Implementation Details

- **Deterministic seeding:** Per-episode RNG seeded from `seed XOR (episode * 0x9E37_79B9_7F4A_7C15)` ensures reproducible randomization across resets while varying across episodes
- **Action latency:** Pending actions stored in a delay line; each substep decrements counters and applies actions whose delay elapsed
- **Observation noise:** Box–Muller gaussian sampling; quaternion perturbations via small-angle axis-angle composition
- **Backward compatibility:** Older kernel WASM builds lacking `resetSeeded` are detected and fail closed rather than silently dropping config; legacy termination check preserved as fallback
- **Reward computation:** Deliberately client-side; kernel always returns 0.0, with `StepInfo` providing all inputs (base pose/velocity, height, tilt, limit violations)

https://claude.ai/code/session_018K8PGkSW8zyrhmVy7Fc3Mw


ecto:
@chojiai the `gain_scale` compounding finding is a false positive — not fixing it, but I've pinned the underlying property with a test.

`PhysicsWorld::set_gain_scale` assigns rather than accumulates (`self.gain_scale = scale`, world.rs:738), so repeated episode draws overwrite each other; there's no `*=` to compound. More importantly, the premise doesn't hold either way: `reset()` rebuilds the world wholesale from `initial_doc` (gym.rs:317) *before* `apply_episode_randomization` runs, so every episode randomizes a pristine model. Nothing carries over between episodes — including `scale_instance_mass` and `scale_joint_friction`, which genuinely *are* multiplicative and would have been the real hazard if the rebuild ever went away.

Since that rebuild is a load-bearing invariant that nothing was guarding, `84bfeea` adds `randomization_does_not_compound_across_episodes`: it seeds episode 0, burns five unrelated episodes, re-seeds the same stream, and asserts an identical rollout. I checked the test isn't vacuous by deleting the rebuild line — joint positions diverge from `[-6.33, -0.77]` to `[-23.76, 66.17]` and it fails.

Worth noting the two nits from the previous round were a good catch in an indirect way: running the physics suite to verify them surfaced a genuine bug of exactly the class described here, just in a different place. `PhysicsWorld::instance_ids()` returned `HashMap` key order, so seeded domain randomization handed the same RNG draws to different links run-to-run, making `seeded_reset_is_reproducible` flaky (2 failures in 15 runs at `5ef2498`). Fixed in `13f39a4` by sorting at the accessor.

---

ecto:
@chojiai good catch on the ambiguity — took the "document explicitly" half of your suggestion in `a317cd2`, not the "compute from the noisy observation" half.

The divergence is the intended contract. `StepInfo` reports the simulator's ground truth; `ObservationNoise` perturbs only `StepResult::observation`. That's the privileged-information split sim2real training relies on — the policy learns against noisy sensors while the reward and the termination decision read true state. Two things would break if `base_height_m` came from the noisy observation: reward would inherit the policy's own sensor noise, and the reported height would no longer match the state `check_termination` actually decided on, so `terminated: true` could arrive next to a `base_height_m` above the threshold.

What was genuinely missing was saying so. The contract is now explicit on `StepInfo`, on the `base_height_m` / `base_tilt_deg` fields, and at the computation site, with a test asserting `info` tracks truth while the returned observation is perturbed away from it — so a future "make these agree" refactor has to break a test rather than silently change the semantics.

77 physics tests, `cargo fmt`, and clippy all pass.

---

ecto:
@chojiai good question to ask — the streams are **shared**, not independent, but episode-level reproducibility survives anyway, and I've pinned the reason in `ea521ce`.

`noisify` draws from the same `self.rng` as domain randomization, so a step does consume draws that would otherwise feed the next episode. What saves it is that `apply_episode_randomization` re-seeds `self.rng` from `seed ^ episode·φ` at the top of *every* reset, discarding whatever the previous episode consumed. So a given `(seed, episode)` pair reproduces regardless of how many steps the previous episode ran.

That makes the re-seed load-bearing for a guarantee nothing was testing, which is the same shape as the `HashMap`-ordering bug earlier in this PR — a reproducibility promise resting on an unguarded implementation detail. So the new test runs two envs, one that took 7 noisy steps and one that took none, re-seeds both to the same stream, and asserts identical randomization and rollouts.

I checked it isn't vacuous by deleting the re-seed: joint positions then diverge from `[-2.68, -2.51]` to `[4.33, 2.84]` and it fails. Worth noting my first attempt at that check was a bad mutation — I gated the re-seed on `episode == 0`, but `reset_with_seed` sets `episode = u64::MAX` and `reset()` pre-increments to 0, so it still re-seeded every time and the test passed green against what I thought was broken code. Deleting the re-seed outright is the mutation that actually exercises it.

78 physics tests, `cargo fmt`, and clippy pass.

---

PR #762, opened 2026-07-31
Add import_urdf MCP tool for URDF robot import

ecto:
Adds a new `import_urdf` MCP tool that imports URDF (Unified Robot Description Format) robot descriptions into session documents, enabling agents to load complete kinematic trees ready for simulation and visualization.

## Summary

The tool parses URDF files via the kernel's WASM parser (the same one the browser drag-drop uses) and builds a full kinematic tree: one part per link, FK-placed instances from joint origins, joints with limits (revolute/prismatic/fixed), and a grounded root link. STL mesh references are resolved against the URDF's directory and optional `package_roots`, parsed server-side, and inlined into the document. Unresolved mesh references fall back to placeholder geometry with explicit warnings, so joint topology and inertials remain accurate even when meshes can't be loaded.

## Key Changes

- **`packages/mcp/src/tools/import-urdf.ts`** — New tool implementation:
  - Accepts URDF via `path` (local filesystem) or `content_base64` (for remote/hosted deployments)
  - Parses binary and ASCII STL files, converting vertex coordinates from meters to mm
  - Resolves `package://` URIs and relative mesh paths against configurable `package_roots`
  - Enforces filesystem jail (`VCAD_MCP_EXPORT_DIR` or cwd) to prevent path traversal
  - Inlines resolved meshes as `ImportedMesh` nodes; reports unresolved references with reasons
  - Registers the imported document as a session and returns kinematic tree summary (parts, joints, instances, ground link)
  - Rejects filesystem paths in remote deployments with clear guidance to use `content_base64`

- **`packages/mcp/src/__tests__/import-urdf.test.ts`** — Comprehensive test suite:
  - Verifies remote mode rejection of filesystem paths
  - Tests import of the 2-DOF arm fixture with session registration
  - Validates `content_base64` mode and unresolved mesh warnings
  - Tests binary STL parsing and unit conversion (meters → mm)

- **`packages/mcp/src/server.ts`** — Registers the new tool in the MCP server

- **`packages/mcp/src/tools/tool-metadata.ts`** — Adds tool metadata entry

- **`packages/mcp/src/__tests__/tool-surface.fixture.json`** — Updates tool surface snapshot with `import_urdf` schema and UI metadata

- **`changelog/entries/2026-07-31-import-urdf-mcp-tool.json`** — Changelog entry

## Notable Implementation Details

- **Mesh resolution mirrors Rust behavior:** The `resolveMeshPath` function implements the same logic as the kernel's `UrdfReadOptions::resolve_mesh`, checking `package://` URIs first, then absolute/relative file:// paths, with filesystem jail enforcement at each step.
- **STL parsing:** Detects binary vs. ASCII format; binary parser reads triangle count and 50-byte records; ASCII parser uses regex to extract vertex coordinates. Both convert from ROS meters to mm (×1000) to match kernel primitive scaling and browser mesh loaders.
- **Size limits:** URDF capped at 8 MB (kernel's own limit), individual meshes at 100 MB to prevent memory exhaustion.
- **Session durability:** Imported documents are registered via `registerSession` and persist according to the server's configured session store (disk or Supabase).
- **Graceful degradation:** Unresolved meshes don't fail the import; links get placeholder cubes, and the tool reports which references couldn't be resolved and why.

https://claude.ai/code/session_01QCLywtbBg6H5HjdHcCKBBW

### DIFF
--- a/changelog/entries/2026-07-31-import-urdf-mcp-tool.json
+++ b/changelog/entries/2026-07-31-import-urdf-mcp-tool.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-31-import-urdf-mcp-tool",
+  "version": "0.9.4",
+  "date": "2026-07-31",
+  "category": "feat",
+  "title": "import_urdf MCP tool",
+  "summary": "Agents can import a URDF robot (links, joints, FK-placed instances, STL meshes) straight into a session document, ready for create_robot_env.",
+  "features": ["urdf", "robotics", "physics", "import"],
+  "mcpTools": ["import_urdf"]
+}
--- a/packages/mcp/src/__tests__/import-urdf.test.ts
+++ b/packages/mcp/src/__tests__/import-urdf.test.ts
@@ -1,0 +1,95 @@
+import { describe, it, expect, afterEach, beforeAll } from "vitest";
+import { importUrdf } from "../tools/import-urdf.js";
+import { getSession, documents } from "../tools/session.js";
+import { Engine } from "@vcad/engine";
+import { mkdtempSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+
+let engine: Engine;
+beforeAll(async () => {
+  engine = await Engine.init();
+});
+
+afterEach(() => {
+  delete process.env.VCAD_MCP_REMOTE;
+  delete process.env.VCAD_MCP_EXPORT_DIR;
+  documents.clear();
+});
+
+const EXAMPLES_DIR = resolve(__dirname, "../../../../examples");
+
+function parseResult(res: { content: Array<{ type: string; text: string }> }) {
+  return JSON.parse(res.content[0].text);
+}
+
+describe("import_urdf", () => {
+  it("rejects filesystem paths in remote mode", () => {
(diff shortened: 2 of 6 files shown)


ecto:
Re: the preview gap in the Choji review summary — I checked this and the preview does fire for `import_urdf`; no change is needed.

`resolvePreviewDocumentId` doesn't need to know the tool by name. Before it reaches the `import_step` / `create_cad_loon` name-specific branches, it tries a generic session-backed branch: if the result's `document_id` parses out and `documents.has()` it, that id is returned directly. `import_urdf` registers the document via `registerSession` and returns that id at the top level of its payload, so it takes the same path `sheet_metal_create` and `open_document` already do.

Verified end to end by driving the real server through an in-memory MCP client (the harness `tool-surface.test.ts` uses) and calling `import_urdf` on `examples/robot-arm-2dof.urdf`. The result carries:

```
_meta keys: [ 'vcad.io/preview', 'io.vcad/build' ]
preview meta: {"document_id":"doc_1_GbWTQ8_-rviOivL_","glb":"<229708 chars>","version":"1icuu4z"}
```

So the handle is attached and the ~230 KB inline first-paint GLB rides along in `_meta` — meaning it's actually the zero-round-trip path, not the `get_preview_glb` fallback the summary describes.

---

PR #767, opened 2026-07-31
feat(physics): warn when PD gains are unstable for the explicit integrator

ecto:
## Why

Gains published for Isaac-style simulators assume **implicit** actuator integration and carry no stability bound with them. This crate integrates explicitly (`phyz::rigid::semi_implicit_euler`), where a PD servo is only stable while ω·dt stays small, ω = √(kp / I_reflected).

Follow booster_gym's published K1 config (kp 200 hip/knee, kd 5, kp 50 ankle) at dt = 5 ms and the light joints land near ω·dt ≈ 0.7 and diverge: base tilt goes 0.1° → 4.65° → 49° in three env steps *while still airborne*. That looks exactly like a physics bug and sends you through the contact solver. It isn't one — raising the tick to 1 kHz (dt = 1/1000, substeps = 20, control still 50 Hz) makes the identical gains stable.

## What changed

- **`PhysicsWorld::check_gain_stability(dt) -> Vec<GainWarning>`** — for each explicitly-gained joint, probes `reflected_inertia`, computes ω·dt, warns above the limit. Sorted worst-first.
- **`GainWarning`** carries `joint_id`, `kp`, `reflected_inertia`, `omega_dt`, `max_stable_kp`, `min_substeps`, and a `Display`:
  > `joint 'crank': kp=200.0000 with reflected inertia 2.216e-5 gives omega*dt = 15.02 (unstable above ~0.3); raise substeps to >= 204x or lower kp below 0.0798`
- **`RobotEnv::check_gain_stability()`** — the timestep is known there, so it supplies `self.dt` and rescales `min_substeps` from a multiplier to an absolute substep count. `RobotEnv::set_joint_gains` emits the warning for the joint just set, which is where the K1 case would have surfaced it.
- **Nothing is clamped.** The caller may be about to raise `substeps`, which is the fix that keeps the published gains intact.

## The threshold is measured, not assumed

`gain_stability_limit_is_conservative` sweeps ω·dt on the existing mm-scale crank fixture in `tests/servo_stability.rs`:

| ω·dt | outcome |
|---|---|
| 0.3 / 0.5 / 0.8 | settles on the 30° target |
| 1.0 | overshoots |
| 1.3 | thrown off target entirely (final −175°) |

So `GAIN_STABILITY_LIMIT = 0.3` sits ~3× under the real boundary — the margin the nonlinear multi-body case needs once coupling, contact impulses, and gain randomization are in play.

## Tests

- `gain_stability_limit_is_conservative` — the calibration sweep above.
- `unstable_gains_warn_and_clear_at_5x_substeps` — gains at ω·dt = 0.9 with 200 Hz substeps produce a warning naming `crank`; the same gains at 1 kHz / 20 substeps (identical control period) produce none.

Full `vcad-kernel-physics` suite passes.

## Reviewer notes

- The test can't use the literal kp = 200: this fixture's reflected inertia is 2.2e-5 kg·m², so kp 200 is ω·dt ≈ 15 at any sane substep and never clears at 5×. It picks inertia-relative gains to reproduce the same regime; the real case is recorded in the docstring.
- `cargo clippy -p vcad-kernel-physics --all-targets` fails on a **pre-existing** `absurd_extreme_comparisons` at `src/gym.rs:1669` (`<= u32::MAX`), present on the base commit and untouched here.
- Velocity-mode servos use `kd` only; the check covers the position-servo `kp` path.


ecto:
Addressed the `gain_scale` nit: `RobotEnv::check_gain_stability` now evaluates at the top of the configured `pd_gain_scale` range instead of the currently-sampled value, so a check run before the first reset can no longer clear gains that go unstable a few episodes later (new test `gain_check_uses_worst_case_randomized_gain_scale`). Left `min_substeps` as-is — `substeps` is fixed for the env's lifetime and is not randomized, so the current value is the correct basis.

PR #769, opened 2026-07-31
feat(urdf): floating-base import option

ecto:
## The problem

Hit for real in #766. Booster's `robots/K1/K1_22dof.urdf` and `K1_locomotion.urdf` — and this is the norm, not a Booster quirk; the in-repo `examples/unitree-g1.urdf` does it too — ship the world link and floating joint commented out:

```xml
<!-- <link name="world"/>
<joint name="world_joint" type="floating">
  <origin xyz="0 0 0"/>
  <parent link="world"/>
  <child link="Trunk"/>
</joint> -->
```

because the convention is that the simulator supplies the floating base (MuJoCo's MJCF for the same robot has `<joint name="world_joint" type="free"/>`). vcad's importer saw no world joint, made `Trunk` the `groundInstanceId`, and welded the robot to the world — physically pinned, useless for locomotion. Training anything meant hand-editing the XML to uncomment that block: exactly the kind of undocumented ritual that costs an afternoon.

## What changed

**Kernel** (`crates/vcad-kernel-urdf`) — `UrdfReadOptions` gains `floating_base`, `floating_base_link`, and `spawn_height_mm`. When set and the URDF declares no floating joint, `apply_floating_base` injects exactly the commented-out block into the parsed `Robot` before the document build: a geometry-less `world` link plus a `type="floating"` joint to the root link. Everything downstream is untouched — the existing `"floating" => JointKind::Free` mapping, root-link discovery, and ground-instance selection do the rest.

- Names uniquify (`world_1`) if `world`/`world_joint` are already taken.
- An unknown `floating_base_link` is an error.
- A URDF that already declares a floating joint is a no-op (the authored origin wins).

**Spawn height** — a `Free` joint's `state` is a scalar and meaningless for 6 DOF, so `spawn_height_mm` is written as the joint's `parentAnchor.z`. That is what actually sets the initial base height; documented at every layer it's exposed. For the K1, 620 mm spawns it just above its 549.8 mm settled standing height.

**Detection** — `commented_out_floating_joint(xml)` scans comment bodies for `<joint type="floating">` and returns its name. The reader warns on stderr when it hits and `floating_base` wasn't requested; `import_urdf` returns it as a `floating_base_warning`. On the real G1 this fires and names `floating_base_joint`.

**CLI** — `vcad import-urdf --floating-base [--floating-base-link LINK] [--spawn-height-mm MM]`. Also added to `vcad simulate`, whose URDF path had the identical trap.

**MCP** — `import_urdf` gains `floating_base` / `floating_base_link` / `spawn_height_mm`, plumbed through two new WASM exports (`importUrdfBufferWithOptions`, `urdfCommentedFloatingJoint`) and `Engine.importUrdf(data, opts?)`. Sub-options without `floating_base` are rejected rather than silently ignored.

## Testing

- 7 reader unit tests: synthesis, default-off, already-declared no-op, explicit link, unknown link, name collision, comment detection.
- `unitree_g1_floating_base_falls_under_gravity` (`crates/vcad-kernel-physics/tests/unitree.rs`) against the in-repo `examples/unitree-g1.urdf`: exactly one `Free` joint at the requested anchor, base falls >1 m in 1 s under gravity — and the **same URDF imported without the flag stays pinned to within 1e-6**, so default behavior is unchanged.
- 4 `import_urdf` tests covering the MCP surface end to end (warning path, synthesis, sub-option rejection, no-false-positive).
- Tool-surface fixture regenerated (`UPDATE_TOOL_SURFACE=1`); the diff is only the new `import_urdf` fields and description.

## Reviewer note

The two MCP floating-base tests need the new WASM exports, so they fail against the *checked-in* `packages/kernel-wasm` artifacts. They pass against a fresh `wasm-pack` build (verified locally, then reverted per the no-committed-artifacts rule in CLAUDE.md); CI's TypeScript job builds WASM from source before running them. Locally, `npm run build -w @vcad/kernel-wasm` first.

Pre-existing clippy findings in `vcad-cli/src/raycast.rs` and `vcad-kernel-physics/src/gym.rs` are unrelated to this branch (confirmed by stashing).

### DIFF
--- a/changelog/entries/2026-07-31-urdf-floating-base.json
+++ b/changelog/entries/2026-07-31-urdf-floating-base.json
@@ -1,0 +1,10 @@
+{
+  "id": "2026-07-31-urdf-floating-base",
+  "version": "0.9.4",
+  "date": "2026-07-31",
+  "category": "feat",
+  "title": "URDF import can synthesize a floating base",
+  "summary": "Humanoid URDFs ship the world link and floating joint commented out; import_urdf's floating_base option injects them so the robot can fall and walk instead of being welded to the world.",
+  "features": ["urdf", "physics", "simulation"],
+  "mcpTools": ["import_urdf"]
+}
--- a/crates/vcad-cli/src/main.rs
+++ b/crates/vcad-cli/src/main.rs
@@ -78,6 +78,25 @@ enum Commands {
         input: PathBuf,
         /// Output .vcad file
         output: PathBuf,
+        /// Synthesize a floating (6-DOF) base when the URDF has none.
+        ///
+        /// Most humanoid/quadruped URDFs ship the world link and its
+        /// `type="floating"` joint commented out, leaving the robot welded
+        /// to the world on import. This injects them.
+        #[arg(long)]
+        floating_base: bool,
+        /// Link to attach the floating base to (default: the tree's root).
+        #[arg(long, value_name = "LINK", requires = "floating_base")]
+        floating_base_link: Option<String>,
+        /// Initial base height in mm for the synthesized floating base.
+        /// Spawn just above the settled standing height (Booster K1: 620).
+        #[arg(
+            long,
+            value_name = "MM",
+            default_value = "0",
+            requires = "floating_base"
+        )]
+        spawn_height_mm: f64,
     },
 
     /// Render document to image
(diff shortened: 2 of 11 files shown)


ecto:
Addressed the parity nit: `vcad simulate` now takes `--floating-base-link <LINK>` alongside `--floating-base` / `--spawn-height-mm`, matching `import-urdf`. Verified on `examples/unitree-g1.urdf` — the synthesized `world_joint` shows up in the sim's joint state dump.

PR #771, opened 2026-08-01
fix(booleans): three band-redesign regressions from #758 (CI 6h hang, bore winding, pinch wedges)

ecto:
## Why

CI had been burning the full **6h GitHub job limit** on every PR since #758 — 12 runner-hours per run across the stable + nightly matrix. Both Rust jobs were killed mid-test, so no PR had a real Rust signal since 2026-07-31 17:22.

One test caused it: `a2_boolean_over_pattern_doc10` went from **2m53s to 4h16m** and started returning wrong geometry. The only kernel commit in that window is `c3c884d1` (#758, chain-based CylBand redesign). Bisecting against its parent `2c251ee0` turned up three distinct defects.

## 1. `parse_band` assumed the loop started on a chain boundary

It decomposed a face's outer loop into its two chains by scanning the vertex list **linearly** (`for i in 0..n-1`), so it never saw the direction change carried by the closing edge `v[n-1] → v[0]`. A loop whose starting half-edge sits mid-chain presents only one visible flip; the decomposition hands the tail a single vertex and bails (`short chains N 1`).

`reversed_winding` in the same function already compensated for exactly this — *"judging by which run came first breaks when the loop's starting half-edge sits mid-chain"* — but the decomposition itself never did.

The damage wasn't one refused face. Each refusal made `split_cylindrical_face_by_sampled` return the face **unsplit**, so the boolean kept geometry it should have trimmed and the next operand split the leftovers again. Fragmentation compounded across the pattern's instances.

**Fix:** when the linear parse fails, re-anchor at each true cyclic turning point and retry. Loops that already start on a boundary take the identical path, so passing cases are untouched.

## 2. `realize_bands` replayed `band.reversed_winding`

Emitting bore-convention loops double-applies an orientation the sewing stage already establishes. On the two-half-annuli union it inverted the r20 bore wall — that wall's area 628 → 890 mm², the solid 7777 → 6031 mm³ (`half_annuli_union_full_annulus`, passing before #758, failing after).

The replay earns nothing anywhere measurable: the 752-case torture corpus scores identically with and without it, as does every other boolean and tessellate test, as does the whole catalogue. Zero demonstrated benefit, one demonstrated regression — so stop replaying it. The field stays for diagnostics with the evidence recorded on it.

## 3. `parse_band` refused collapsed-pinch wedges

Where a band's two chains meet at a shared vertex, sewing leaves a **three**-vertex loop. `try_parse_two_chain` can't express the resulting one-vertex chain and the `n < 4` floor refused it outright, so every such wedge came back unsplit. Built directly instead — the shape is one the downstream machinery already expects.

## Results

| doc10 (23 blades) | before | after |
|---|---|---|
| volume (expected 3365) | 6295 (+87%) | **3346** (−0.6%) |
| faces | 1856 | 593 |
| catalogue binary | 6h kill | **17.6s** |

## Also here

- **Cache the `VCAD_BAND_DEBUG` lookup** — the same bug #758 fixed for `VCAD_SPLIT_DEBUG` and missed here. `band_dbg!` sits inside `try_parse_two_chain`, which runs per-face and up to four times per face, and `env::var` allocates and takes the environment lock on every call.
- **Report which stage refused a sampled split.** The old message always blamed the intersection curve (*"not a closed single-valued profile"*) even when the curve was fine and the FACE was rejected. Every failure here was the latter — that misdirection is most of why this was hard to find.
- **Cap the Rust CI jobs at 75 minutes.** There was no `timeout-minutes` on any job, which is why one slow test could cost 12 runner-hours.

## Testing

- `vcad-kernel-booleans` + `vcad-kernel-tessellate`: green, 126 tests, including `half_annuli_union_full_annulus`.
- Torture corpus: 643 pass, **no regressions**, +2 improvements (rand-052, rand-180 — not baselined; the committed baseline is x86_64, this ran on aarch64).
- `torr_boolean_catalogue`: 24 passed, 2 failed, 6 ignored — completes instead of hanging.
- `clippy -D warnings` clean, `fmt` applied.

## Still failing — reviewer please note

**This does not make CI green.** Two catalogue cases remain, both pre-existing on main:

1. **doc10 undershoots by 19.4 mm³ against a ±16 tolerance** — about 0.84 mm³ per blade, the size of the chord-vs-arc sliver lost where each blade meets the r45 wall at the boolean's segment count while the oracle integrates the true cylinder at 256. A discretization gap, no longer a topology defect.

2. **`c2_extrude_union_extrude_half_annuli` reads +64%. Localized but not fixed.** The culprit is `weld_boundary_vertices` in `repair.rs`, which merges boundary vertices at 1.5e-3 without checking whether the two are already joined by an edge. Welding across such a pair doesn't join two boundaries, it *collapses* that edge — and on a thin overlap wedge it eats the wedge. With the weld off the union reads 7777.29 vs 7854 expected, matching pre-758 exactly, and operand order stops mattering (12861/12886 → 7777 both ways).

   The obvious guard — skip pairs already joined by a half-edge — is **wrong**: it costs 7 torture crashes (`half-edge has no next`) and 3 unit failures, so some adjacent welds are load-bearing. Tried and reverted. A correct fix has to separate "genuinely degenerate edge" from "thin feature". Note the eval path is also wrong at 0° and 135° overlap, which the weld does not explain — a self-union should be idempotent and reads 5376 vs 4104.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,12 @@ jobs:
   rust:
     name: Rust (${{ matrix.toolchain }})
     runs-on: ubuntu-latest
+    # A single pathological test (a kernel perf regression, a hang) would
+    # otherwise burn the full 6h GitHub job limit — 12 runner-hours per PR
+    # across the matrix — before anyone learns anything. The whole job runs
+    # in ~35m healthy, so cap it well below the platform ceiling and let a
+    # blowup surface as a fast red.
+    timeout-minutes: 75
     # Nightly clippy adds new lints constantly; treat it as informational.
     continue-on-error: ${{ matrix.toolchain == 'nightly' }}
     strategy:
--- a/changelog/entries/2026-08-01-boolean-band-cyclic-parse.json
+++ b/changelog/entries/2026-08-01-boolean-band-cyclic-parse.json
@@ -1,0 +1,9 @@
+{
+  "id": "2026-08-01-boolean-band-cyclic-parse",
+  "version": "0.9.4",
+  "date": "2026-08-01",
+  "category": "fix",
+  "title": "Booleans: parse cylindrical bands independent of loop start",
+  "summary": "Booleans over many-instance patterns no longer fragment faces or drop geometry when a face loop's starting edge sits mid-chain.",
+  "features": ["booleans", "patterns"]
+}
--- a/crates/vcad-kernel-booleans/src/cyl_band.rs
+++ b/crates/vcad-kernel-booleans/src/cyl_band.rs
@@ -78,9 +78,19 @@ pub(crate) struct CylProfile {
     vs: Vec<f64>,
 }
 
+/// Whether `VCAD_BAND_DEBUG` is set, read once per process. `band_dbg!` sits
+/// inside `try_parse_two_chain`, which runs per-face and up to four times per
+/// face (once per tie policy), so a per-call `env::var` — which allocates and
+/// takes the environment lock — is not affordable here. Mirrors
+/// `split::split_debug_enabled`.
+fn band_debug_enabled() -> bool {
+    static ON: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    *ON.get_or_init(|| std::env::var("VCAD_BAND_DEBUG").is_ok())
+}
+
 macro_rules! band_dbg {
     ($($arg:tt)*) => {
-        if std::env::var("VCAD_BAND_DEBUG").is_ok() {
+        if crate::cyl_band::band_debug_enabled() {
             eprintln!($($arg)*);
         }
     };
@@ -329,6 +339,13 @@ pub(crate) fn parse_band(brep: &BRepSolid, face_id: FaceId) -> Option<(CylinderS
         .collect();
     // Faces with holes are outside this family.
     if !face.inner_loops.is_empty() {
+        if crate::split::split_debug_enabled() {
+            eprintln!(
+                "parse_band: {face_id:?} rejected — {} inner loop(s), {} outer verts",
+                face.inner_loops.len(),
+                verts.len()
+            );
+        }
         return None;
     }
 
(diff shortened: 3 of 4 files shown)


ecto:
@chojiai Addressed the turning-point nit in f574cd70.

You were right that it's harmless, and for the reason you gave — anchoring at vertex 0 is the identity rotation the linear pass already tried. I made that explicit rather than incidental: the retry loop now skips `start == 0`, which also saves four parse attempts per face on every loop whose vertex 0 happens to be a turning point (this runs per-face, so it's worth having).

Also documented the broader `leaving == 0` skip, which is the more interesting half. Skipping those vertices is intentional for all indices, not just 0: a vertex with a vertical outgoing step is the foot of a connector, and anchoring there would open the rotated list with the connector instead of a chain. The vertex at the connector's far end — the first that actually leaves in the reversed direction — is the anchor the parse wants, and that one is reported. It mirrors the existing `split_at` walk-back, which likewise pushes connector vertices onto the chain that follows them.

No behavior change beyond the dropped redundant parses: booleans + tessellate green, torture 643 pass / no regressions, catalogue byte-identical (doc10 3345.944, 17.2s).

On the sandbox check failures — agreed those are environmental. `cargo check` needs the sibling `tang` and `phyz` repos cloned next to the checkout, and the JS builds need `wasm-pack`; both are documented prerequisites in CLAUDE.md.

ecto:
CI results are in, and they match the local runs exactly — no surprises.

| | before | after |
|---|---|---|
| Rust (stable) | 6h kill, no signal | **fail in 36m36s** |
| Rust (nightly) | 6h kill, no signal | **fail in 38m56s** |
| torr catalogue (CI debug build) | 4h16m → never finished | **28m40s** |
| Torture track (full, x86_64) | — | ✅ **pass, no regressions** |

Both Rust jobs report `24 passed; 2 failed; 6 ignored` — the two known cases, nothing nightly-specific. Everything else green: WASM build, TypeScript, KiCad, both audits, workspace `clippy -D warnings`, `fmt`.

The torture pass on x86_64 is the meaningful confirmation: my local run was aarch64 against an x86_64 baseline, so its "+2 improvements" weren't trustworthy. CI ran it on the baseline's own platform — clean across all 752 cases.

### One more negative result on the c2 weld defect

I took a second run at `weld_boundary_vertices` after the first (refuse all adjacent pairs) failed. The theory was that the coarse 1.5e-3 seam-weld radius can't distinguish a genuinely degenerate edge from a thin real feature, so adjacent pairs should be judged at the caller's *fine* tolerance instead — degenerate ones still weld, real thin features survive.

**It fails identically** — same 7 torture crashes (`half-edge has no next`), same 3 unit failures, same cases. Reverted.

That rules out the whole tolerance-tuning family: the load-bearing welds live in the **same tolerance band** as the wedge-destroying ones, so no threshold separates them. A correct fix needs a different discriminator entirely — something topological (would collapsing this edge eliminate real face area? is this loop a thin wedge or a degenerate stub?) rather than metric.

Also still unexplained and not the weld: the eval path is wrong at 0° overlap, where a self-union should be idempotent and reads 5376 vs 4104. Likely a fourth defect in the same redesign.

### Where this leaves the PR

Three real regressions fixed and the 6h burn stopped, but **CI is not green and this PR doesn't claim it will be**. The two remaining failures are documented in the body with reproductions. Merging trades a 12-runner-hour-per-PR blackout for a 36-minute red — worth it on its own, but the catalogue stays red until someone lands the weld fix.

PR #772, opened 2026-08-01
feat(sim): GPU batch path — real model, PD position targets, ARS step decay, fail-closed base termination

ecto:
Follow-ups to #770 (the K1 standing solve), in two groups: trainer/env hardening, and the GPU batch path bring-up.

## Trainer + env hardening

**`ArsConfig::step_size_final`** — optional cosine step-size decay. The ARS update is scale-free, so at any *constant* alpha the trainer keeps taking full-size steps after it has found a solution; shrinking the constant only slows the wandering. Ships **opt-in, default `None`**, because the measurement says it doesn't help here:

| arm | seed 7 | seed 23 |
|---|---|---|
| constant alpha = 0.005 | 395.01, 10/10 | 395.12, 10/10 |
| cosine decay to 0.0005 | 377.51, 10/10 | 394.45, 10/10 |

Decay never wins and hurts seed 7 — annealing bites just as the curriculum reaches full randomization and the policy still needs step authority. Held-out-best selection already neutralizes the wandering, and the task is saturated (~395 against a ~400 ceiling), so there is no headroom for a schedule to buy. The knob stays because it will matter on a task with a real optimum to settle into.

That matrix also closes the open question from #770: **a second training seed reproduces 10/10** (395.12 vs the original 386.92), so the solved-task claim is not a lucky draw.

**Fail-closed base termination.** `RobotEnv::new_with_config` now errors (`PhysicsError::UnobservableBase`) when `base_height_below` / `base_tilt_above_deg` are configured but no base pose is observable. A fixed-base document or a typo'd `base_instance_id` previously disabled *every* termination check silently, so a run reported a confident 400/400 survival while measuring nothing at all. The `k1_stand` example had this guard privately; every `RobotEnv` caller now gets it.

## GPU batch path

**The batch pipeline now runs the real physics model.** `BatchSimPipeline::from_document` builds through `PhysicsWorld::from_document` and clones its model — authored inertials, collider masses, joint frames, limits. It previously re-derived the model with density-1000 box inertias, i.e. it trained against a different robot than the CPU env simulated. Same family as the placeholder-cube bug.

**PD position targets on device.** `enable_pd` maps every single-DOF vcad joint onto the new `phyz-gpu` servo pass (ecto/phyz#35) with per-joint gains and the authored effort limit as the torque clamp; `batch_step_targets` is the rollout hot path, so only targets cross the bus. `PhysicsWorld::joint_addressing` exposes each joint's `(q offset, v offset, ndof, effort limit)` so batched backends index the flat state without re-deriving the layout.

**Measured on the K1** (M2 Mac, Metal, 22 DOF + floating base):

| envs | submit substeps/s | readback substeps/s | vs CPU |
|---|---|---|---|
| 64 | 38.5k | 24.8k | 18x |
| 256 | 153.5k | 97.6k | 72x |
| 1024 | 614k | 358k | 289x |
| 4096 | **2.17M** | 1.07M | 1022x |

CPU reference is one `RobotEnv` at the exact ARS rollout configuration: 2.1k substeps/s. The 4096-env submit column is ~85x the whole 12-core machine's rayon-parallel throughput. The submit/readback gap is the argument for the zero-copy tensor contract (`phyz_gpu::GpuBatchSimulator::interop`).

Sanity checks in `k1_gpu_bench`: a zero-torque K1 falls (gravity + floating base through the batched ABA shader), and all 22 servoed joints track a 0.3 rad command to <1e-4 while the base free-falls. The PD check deliberately commands a *nonzero* target — a zero-target version passes vacuously, since uniform free fall holds the joints at rest with or without servos.

## Known gaps (next milestone)

- No collision geometry on the model, so the GPU penalty-contact pipeline has nothing to touch. This is the last blocker to actually training on the GPU path.
- f32 drift over full episodes is unquantified; `phyz-env::BatchEnv` is the f64 CPU oracle to assert against.
- One submit per substep — folding the substep loop into the shader multiplies the submit column again.

Depends on ecto/phyz#35 for the PD pass.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelog/entries/2026-08-01-ars-step-decay-failclosed-base.json
+++ b/changelog/entries/2026-08-01-ars-step-decay-failclosed-base.json
@@ -1,0 +1,9 @@
+{
+  "id": "2026-08-01-ars-step-decay-failclosed-base",
+  "version": "0.9.4",
+  "date": "2026-08-01",
+  "category": "feat",
+  "title": "ARS step-size decay + fail-closed base termination",
+  "summary": "The RL trainer gains an optional cosine step-size schedule, and robot envs now refuse base-pose termination conditions that could never fire (fixed-base or mistyped base instance).",
+  "features": ["simulation", "rl", "physics"]
+}
--- a/crates/vcad-kernel-physics/src/error.rs
+++ b/crates/vcad-kernel-physics/src/error.rs
@@ -38,6 +38,23 @@ pub enum PhysicsError {
     #[error("No ground instance specified in document")]
     NoGroundInstance,
 
+    /// Termination conditions reference a base pose no instance provides.
+    ///
+    /// Fail-closed guard: with `base_height_below` / `base_tilt_above_deg`
+    /// configured but no observable base pose, those checks would silently
+    /// never fire and every episode would run to `max_steps` — a run that
+    /// reports confident survival while measuring nothing.
+    #[error(
+        "termination config sets base-pose conditions, but base instance {base_instance_id:?} \
+         has no observable pose — the checks would silently never fire. Point \
+         `EnvConfig::base_instance_id` at the floating-base instance (or import the \
+         floating-base variant of the robot)"
+    )]
+    UnobservableBase {
+        /// The configured (or defaulted) base instance id, if any.
+        base_instance_id: Option<String>,
+    },
+
     /// Evaluation error.
     #[error("Failed to evaluate geometry: {0}")]
     Evaluation(String),
--- a/crates/vcad-kernel-physics/src/gym.rs
+++ b/crates/vcad-kernel-physics/src/gym.rs
@@ -363,6 +363,24 @@ impl RobotEnv {
             ground,
             joint_gains: std::collections::HashMap::new(),
         };
+        // Fail closed on base-pose termination conditions that can never
+        // fire. A fixed-base document (or a typo'd `base_instance_id`) makes
+        // `Observation::base_pose` None on every step, which silently
+        // disables `base_height_below` / `base_tilt_above_deg` — the episode
+        // then always runs to `max_steps` and an RL run reports confident
+        // survival while measuring nothing. That mistake has cost a real
+        // afternoon of training; it should be a constructor error, not a
+        // property of the numbers.
+        let wants_base_termination = env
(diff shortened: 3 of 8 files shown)


ecto:
Pushed `a4a933ac`: GPU contact now works and **the K1 stands on the GPU ground plane** — base settles 0.111 m below spawn and holds, against a 4.3 m free fall.

Getting there turned up three things worth calling out, two of them bugs that were silently wrong rather than loudly broken.

**GPU ground contact had never worked at all.** Not a K1 problem, not a vcad problem: a single free box dropped from 1 m reached 1e17 m in 400 steps. The contact shader's FK world accumulation was inverted, so a body *above* the plane read as below it and the penalty force pushed it into deeper apparent penetration. Damping and friction were accepted as parameters and silently ignored. Support points assumed an axis-aligned body. All fixed upstream in ecto/phyz#35, with a dropped box now settling at exactly the analytic depth `g*tc^2`.

**The batch pipeline was running 4.2x coarser physics than the CPU env.** `PhysicsWorld` builds its model with `dt = 1/240`, but `RobotEnv` overrides that on every step — the RL config runs 1/1000 with 20 substeps. The GPU simulator bakes `model.dt` into a uniform at construction, so every GPU number to date was taken at the wrong timestep, with PD gains chosen for the finer tick. `from_document_with_dt` makes it explicit and the bench now passes 1 kHz.

**The document's ground plate needed stripping.** `prepare_colliders_for_gpu` now drops geometry from immovable bodies (the rule `PhysicsWorld` already applies to its own contact set) and box-approximates the rest. Without the strip, the ground plate sits at z <= 0 and the penalty shader sees a metre-deep penetration on step one.

Two honest caveats on this milestone:

- The GPU settles ~0.11 m lower than the CPU env's 0.5498 m. Contact *works*; contact does not yet *agree*. Chasing that is M3, against `phyz-env::BatchEnv` as the f64 oracle.
- Box feet are a real approximation. Conservative (a foot contacts no later than its mesh would), but a policy trained through this path has box feet, and that is a sim2real fact rather than a rounding error.

Throughput is unchanged by the contact work: 2.18M substeps/s at 4096 envs.

PR #776, opened 2026-08-02
fix(ci): green the Rust job — 303 test binaries in 9.6 min, up from 38 in 22.7

ecto:
## The problem, and why it kept moving

`cargo test --workspace` was red on main. It looked like "two failing tests" because **cargo stops at the first failing test *binary*** — CI only ever reached 38 of ~300 binaries, so each fix revealed the next failure instead of turning the job green. (#774 quarantined two; main went red again on a third.)

A `--no-fail-fast` sweep found the whole set at once: **7 failures, 3678 passing**. This PR resolves all 7, fixes an unrelated upstream break, and adds `--no-fail-fast` so the suite can never hide a queue again.

## Measured result on CI

| | before | after |
|---|---|---|
| test binaries actually run | 38 (job bailed early) | **303** |
| test execution | 22.7 min | **9.6 min** |
| `torr_boolean_catalogue` | 20.8 min | **0.89 min** |
| status | red | **green** |

8x the tests actually executed, in 42% of the time.

## Four were stale tests, not kernel bugs — fixed properly

#758 moved boolean-result rims onto the sag-adaptive canonical grid (`split.rs::arc_segments`), so a rim discretizes at `max(requested, sag_count)` points — **112 for r=2.5, not `circle_segments = 32`**. Four differentiable-seam tests still hardcoded the inscribed 32-gon.

Recomputing with the kernel's actual rim count reproduces the measured volume **exactly**:

```
closed form with n=112: 301.8767173349144
kernel mesh volume    : 301.8767173349144
relative error        : 1.9e-16     (test gate: 1e-12)
```

The kernel is also strictly *more* accurate now (hole area 19.6247 vs the true circle's 19.6350; the old 32-gon gave 19.5090). `arc_segments` is now public and documented, and the tests ask it rather than pinning their own copy of the convention: `m2_through_hole_dv_dr_matches_closed_form_and_fd`, `m2_rim_nodes_implicit_diff_matches_fd`, `m10_hole_second_derivative_matches_closed_form_and_fd`, `m2_hole_radius_synthesized`.

## Three are real #758 defects — quarantined, not fixed

**`control_static_group_volume`** — the useful one. Reduced to a **two-solid repro**: `base ∪ stator1` = 38313.784 vs 42235.415 exact (**−3921.631**), where base tops out at z=25 and stator1 spans z∈[26, 28.7] — *the two solids do not touch*. A union of Z-disjoint solids is losing volume. Every leaf measures correct to <1.1 mm³, and in the full tree the entire error appears at that one step and stays flat across 6 later unions — one bad operand pair, not accumulation.

**`test_flanged_hub_mesh_validation`** — 1 degenerate triangle (expected 0); not diagnosed past the count.

**`through_hole_reconstructs_full_circles`** — STEP exports 6 `CIRCLE` entities instead of 8; export fidelity, not a volume error.

## An unrelated upstream break, also fixed

Partway through, **every** Rust job (stable, nightly, WASM) started failing to compile: phyz main added a `dt` parameter to `phyz::contact::assemble` (ecto/phyz#36-38) and vcad's call site still passed 6 arguments. CI clones phyz from main **unpinned**, so this landed on vcad main with no vcad commit.

Passing `dt` alone would double-stabilize — vcad applied its own capped Baumgarte push right after `assemble`, which is exactly what phyz's solref bias now does per contact row (~9%/step at dt=1/240 vs vcad's 20% of depth-minus-slop). Upstream owns this now, so the hand-rolled bias is removed. Verified by the 109 physics tests including `contact_observation`, `servo_stability` and `m11_contact_gradient`.

Worth a follow-up: an unpinned sibling clone means any phyz merge can redden vcad CI.

## Speed: what changed and what it cost

Tests ran in the dev profile at **opt-level 0** — brutal for a numerics kernel. Same 24 tests, same machine: **158.4 s → 7.9 s** at opt-level 2, and on CI the catalogue went 20.8 min → 0.89 min. Unlike `--release` this keeps `debug_assertions` and overflow checks. It generalizes the `[profile.dev.package.vcad-kernel-fea]` override already in the file.

The honest tradeoff: optimized builds compile slower, and that is now the bottleneck — a cold-cache run is **21.5 min compiling vs 9.6 min testing**. So this also sets `cache-workspace-crates: true` on rust-cache, which the wasm job already uses; the ~90 vcad crates were being rebuilt every run. Total job time on the cold-cache run was 33 min; the warm-cache steady state should be well below that.

`m3_flywheel_optimize` is marked a perf canary. It **passes** — but 250 L-BFGS iterations x 5 booleans is ~13 min locally, and `--no-fail-fast` means CI would now reach it for the first time. Revert the `#[ignore]` if you'd rather pay it.

## Verification

Full workspace exactly as CI runs it: **3926 passed, 0 failed, 61 ignored**. `cargo fmt --all --check` clean. All 16 CI checks green, including the torture track (which byte-compares against an x86_64 baseline — good evidence `opt-level = 2` does not perturb kernel geometry).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,16 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
           components: rustfmt, clippy
+      # cache-workspace-crates: the ~90 vcad crates are path deps of each
+      # other, and rust-cache leaves those out by default — so every run
+      # rebuilt the whole workspace even when a PR touched one crate. That is
+      # now the dominant cost: with tests running optimized, a cold-cache run
+      # spends 21.5 min compiling and 9.6 min actually running tests. Most
+      # crates are untouched by any given PR, so cache them too (the `wasm`
+      # job below already does this).
       - uses: Swatinem/rust-cache@v2
+        with:
+          cache-workspace-crates: true
       # `cargo fmt --all` would traverse into the path-dep workspaces
       # (tang/phyz/loon, cloned above). Format only vcad workspace members
       # explicitly so external repos don't gate our CI.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,12 @@ jobs:
             | python3 -c "import json,sys; d=json.load(sys.stdin); print('\n'.join(p['name'] for p in d['packages']))" \
             | xargs -n 1 -I{} cargo fmt -p {} --check
       - run: cargo clippy --workspace --exclude vcad-desktop --features vcad-kernel-text/no-builtin-font -- -D warnings
-      - run: cargo test --workspace --exclude vcad-desktop --features vcad-kernel-text/no-builtin-font
+      # --no-fail-fast: `cargo test` otherwise stops at the FIRST failing test
+      # binary, so a single red crate hides every failure behind it. That is how
+      # 5 broken tests sat undetected on main behind the torr catalogue — CI
+      # only ever reached 38 of ~200 binaries, and each fix revealed one more
+      # failure instead of turning the job green. Run everything, report everything.
+      - run: cargo test --workspace --exclude vcad-desktop --features vcad-kernel-text/no-builtin-font --no-fail-fast
       - run: cargo build --workspace --exclude vcad-desktop --examples --features vcad-kernel-text/no-builtin-font
       - run: cargo doc --workspace --exclude vcad-desktop --no-deps --features vcad-kernel-text/no-builtin-font
       # packages/ir/src/generated.ts is generated from the Rust `vcad-ir` types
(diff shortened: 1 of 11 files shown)


ecto:
Fixed in 7b609f6c — good catch, and it was worse than a leftover fragment.

Two stale blocks were stacked above `arc_segments`: an orphaned "Compute the U parameter for a point on a cylinder surface" line, and a pre-#758 contract asserting the count "must stay at the caller's count — sag-adaptive densification here reopens every analytic/frozen seam" — which the code flatly contradicts, since it returns `max(requested, sag_count)`. Making the function public would have published both to rustdoc.

Rewritten as one block. I kept the old warning as a caveat rather than deleting it: the display tessellator re-samples surviving *analytic* circle boundaries at exactly the caller's `n`, so densifying makes a frozen rim disagree with an analytic neighbour. That is a plausible (unproven) cause of the quarantined `through_hole_reconstructs_full_circles` — STEP finding 6 circles where it should find 8 — so it is now recorded as the first place to look.

PR #784, opened 2026-08-08
feat(fea): M4 — discrete shape adjoint for structural response

ecto:
## What

Adds a **discrete shape adjoint** to `vcad-kernel-fea` (M4): `dJ/d(node coordinates)` for a static linear-elastic solve, at the cost of one extra linear solve. The forward solver already answered *will this bracket break*; this answers *which way do I move the geometry to fix it*.

Three objectives (`Qoi`):

| QoI | Units | Adjoint cost | `∂J/∂x|_u` |
|---|---|---|---|
| `Compliance` | N·mm | self-adjoint (`λ = u/E`), **no solve** | 0 |
| `MeanDisplacement { region, direction }` | mm | one solve | 0 |
| `SmoothMaxVonMises { p, threshold_mpa }` | MPa | one solve | non-zero |

`K` is symmetric, so the adjoint reuses the forward operator, preconditioner, and PCG. That asymmetry is the point: finite differences cost one solve *per parameter*, so a 200-parameter part costs 400 solves for a worse answer.

## Why

This is the structural link in pond's co-design chain (`POND.md` §4b). The thesis needs a task-performance gradient that reaches geometry; a gradient that reaches geometry *without* a stress constraint produces parts that are confidently wrong. This is that constraint, and it is differentiable.

**Placement — vcad, not phyz.** Solid mechanics is physics, which argues for phyz. It lost to three practical facts: the solver consumes BRep tessellation, material assignment, and the receipt schema (all vcad's); its output feeds DFM and design verdicts rather than simulation rollouts; and phyz's role in the loop is *upstream*, producing load cases (joint reactions from a rollout) that this crate consumes. phyz never appears inside a static solve. Cost of being wrong is a crate move plus load-case re-plumbing — days, because the adjoint math and the seam contract are placement-independent.

**The tang constraint is satisfied the house way.** Nobody tapes through a 2000-iteration PCG loop; instead this exposes a vector-Jacobian product that contracts with `vcad-kernel-diff`'s tang-taped `dx/dθ`. Same pattern as `vcad-kernel-thermal`'s one-extra-solve adjoint, `vcad-kernel-particle`'s Poisson adjoint, and `phyz::diff`'s contact adjoint. The chain stays unbroken because the seam is where the tape lives, not the solver.

## The bug the numbers caught

The first working version would have shipped a real defect. The **unthresholded** stress p-norm reads **150 MPa against a true peak of 72** on an 80-cell cantilever — every one of 38,400 elements counts as "active", so the bracket factor `N^(1/p)` is worthless. As a yield constraint that rejects safe designs.

Adding the thermal crate's excess-over-threshold treatment:

| p = 8 | J (MPa) | true peak | over-read | active elements |
|---|---|---|---|---|
| no threshold | 150.21 | 71.97 | **+108.7%** | 38,400 |
| τ = 55 MPa | 75.62 | 71.97 | **+5.1%** | 897 |

## Validation

Adjoint vs central finite differences on frozen connectivity, PCG at 1e-12, two velocity fields per QoI (a uniform thickness scale, and a taper that varies along the beam so the full per-node gradient is exercised rather than one global scale):

| QoI | thickness | taper |
|---|---|---|
| compliance | 2.1e-9 | 2.1e-9 |
| tip deflection | 2.1e-9 | 2.1e-9 |
| smooth-max stress (τ=0) | 1.9e-9 | 6.5e-9 |
| smooth-max stress (τ=55) | 2.5e-9 | 3.2e-8 |

Passing tests aren't evidence on their own, so two more things:

**Mutation testing** — three deliberate corruptions of the adjoint, each caught:

| mutation | caught by | error |
|---|---|---|
| drop the explicit `∂J/∂x` term | stress tests **only** | 3.4% |
| drop the test-function term | all four FD tests | 33–49% |
| transpose the stress/gradient pairing | all five gradient tests | 10³–10⁶× |

The selectivity of the first is the useful signal: it perturbs only the objective that has an explicit shape term, exactly as the math predicts.

**A closed-form check**, because FD only proves the adjoint matches the *discrete model* — it cannot catch a discrete model that is wrong. Beam theory gives `d(deflection)/dt = −3·δ_bend/t`; the discrete gradient converges onto it monotonically under refinement (−0.198 at h=2mm, −0.211 at h=1mm, vs −0.217 exact), the residual gap tracking the constant-strain-tet bending stiffness the forward solve already reports.

Also holding: the compliance gradient equals `P ×` the tip-deflection gradient for the single-load case (1e-8); the self-adjoint shortcut agrees with the general path; the p-norm brackets the hard max and tightens monotonically in `p`; the reported `Solution` is the one that was differentiated, not a second solve.

## Reviewer notes

- **`solve.rs` refactor is the substantive non-adjoint change.** Element assembly, the operator, PCG, and `summarize()` are now `pub(crate)` and shared, so the adjoint differentiates *exactly* the system that ships rather than a parallel reimplementation, and a gradient call reports the solution it differentiated instead of re-solving. Public API is unchanged; all 34 pre-existing tests pass untouched. `GradError` is a new type rather than new variants on `SolveError`, so no public enum changed shape.
- **Scope is fixed-topology on a frozen discretization, deliberately.** `tet_fill` re-voxelizes per geometry, so a parameter sweep pops whole tets in and out and `J` is a step function at those events — re-meshing per optimizer step and finite-differencing across it measures the staircase, not the physics. Consumers mesh once and contract with node velocities. Region membership and prescribed nodal loads (`∂f/∂x = 0`) are frozen for the same reason. All three are documented at the module head and in `docs/fea-m4.md`.
- **Commit `c474ffa5` adds a spike example that `56b596de` then deletes.** That is intentional, not churn: the spike was the evidence that the capability was worth building, and the library supersedes it. Its unique contribution (the closed-form gradient check) was moved into the library tests first; a parallel copy of the element math would only drift.
- **No changelog entry, deliberately.** Per `CLAUDE.md` the changelog tracks user-facing behavior, and there is no MCP tool or WASM binding yet — an entry would claim something nobody can do. That wiring is the next chunk.
- **Milestone numbering:** this is M4. The main checkout is currently on an older commit whose `docs/fea-m0.md` still says "M3+ out of scope"; on this branch M3 (thin walls) is already complete.
- `cargo test -p vcad-kernel-fea`: 51 pass. clippy `-D warnings` clean, `cargo fmt --all --check` clean, and both downstream consumers (`vcad-kernel`, `vcad-kernel-wasm`) build.

## Next

The open step that actually closes the loop to loon is **seam registration with `vcad-kernel-diff`** — turning a named `.vcad` parameter into the node velocities this contracts with. After that, phyz load-case ingestion (joint reactions → `FeaSpec` loads) is what makes the capability load-bearing for co-design rather than merely available.

Reproduce the numbers with:

    cargo run --release -p vcad-kernel-fea --example gradient_report


ecto:
Note on the red `Rust (nightly)` check — **this failure is pre-existing on `main` and unrelated to this PR.**

Exactly one test fails workspace-wide (`--no-fail-fast`, so nothing is hidden behind it):

```
fit_physics::tests::cube_in_void_falls_under_gravity
mecheval/graders/src/fit_physics.rs:715: expected ~1.226m of free fall, got 0m
test result: FAILED. 60 passed; 1 failed
```

Evidence it is not caused by this branch:

- **Same failure on `main`s own CI**, at c0da0d6 (2026-08-05, three days before this branch existed) — identical test, file, line, and 60-passed/1-failed count, on [nightly](https://github.com/ecto/vcad/actions/runs/31004913966) *and* stable.
- **Reproduces at this PRs merge-base** (662e4dd6) with the branch checked out and my commits absent.
- **No dependency path**: `mecheval-grader` does not depend on `vcad-kernel-fea`. This diff touches only `crates/vcad-kernel-fea/**` and `docs/**`.
- **The FEA tests themselves pass in the same CI run** (51 unit + 8 doc-tests), as do WASM build, cargo audit, KiCad export, the torture track, and the WASM-artifact guard.

The bug looks real and worth fixing — a body in free fall reporting `0m` drift means it never moved — but it is a phyz/mecheval physics issue in a different subsystem, so fixing it here would balloon this PRs scope and couple an FEA change to a physics regression. Recommend it be tracked separately; happy to take it as its own PR.

ecto:
Fix for the two red Rust checks is up as #785, stacked on this branch.

Root cause: the pinned phyz (rev `7c8ac59`) reordered the free-joint `q` layout to `[wx, wy, wz, x, y, z]`, and `mecheval`s `simulate_drop` still hand-integrated the old translation-first layout — so linear acceleration went into the yaw exponential coordinate and a body released under gravity never fell. phyzs own `joint.rs` documents both the reorder and that exact symptom. Fixed by calling phyzs `integrate_configuration` instead of re-deriving the layout.

With #785 the exact CI command passes: 302 test binaries, zero failures.

One thing worth reading in #785 before merging: restoring motion unmasks a genuine contact defect that the frozen bodies were hiding (`cube_on_plate_settles_under_gravity` was passing vacuously). That test is now `#[ignore]` with the measurements recorded rather than retuned, and `check_gravity_hold` cannot grade resting contact until it is fixed.

PR #786, opened 2026-08-08
feat(sim): simulate XLeRobot, and fix the mesh ops that evaluated to nothing

ecto:
XLeRobot — an IKEA RÅSKOG cart carrying two 6-DOF SO-101 arms and a pan/tilt head — now imports, renders and simulates in vcad.

| | |
|---|---|
| Actuated DOF | **17** (3 base + 12 arm + 2 head) |
| Observation | 86 features |
| Speed | ~20k steps/s (dt 1/200 s, 4 substeps) |
| Geometry | 106,480 triangles |
| Asset | `third_party/xlerobot`, Apache-2.0, pinned `3d14695`, 29 meshes / 4.8 MB |

```bash
cargo run --release -p vcad-kernel-physics --example xlerobot_probe -- examples/xlerobot.vcad
cargo run --release -p vcad-sim --example xlerobot_reach -- examples/xlerobot.vcad
```

It is the counterpart to the Booster K1 rather than a second copy: the K1 is a floating-base humanoid whose whole problem is not falling over, while XLeRobot's base is a *planar* chain (prismatic X, prismatic Y, continuous yaw) on a world-welded root. It cannot tip and never touches the ground collider. Manipulation is the point.

### `MeshImport` evaluated to no solid — the root of a whole bug family

`CsgOp::MeshImport` and `CsgOp::ImportedMesh` both returned `Ok(None)` from the BRep evaluator. `None` means "this subtree contributes no geometry" and it **propagates**, so a mesh under a `Translate` — what a URDF `<visual>` with an `origin` offset imports as — made the *entire part* evaluate to nothing. Downstream: empty renders, zero volumes, booleans quietly dropping a mesh operand, and a hard `no resolvable geometry` failure in physics for any link without an authored `<inertial>`. Every consumer had to special-case a bare mesh node at a part's root to see anything at all.

`Solid` has had a `SolidRepr::Mesh` variant and `Solid::from_mesh` all along — transforms (including winding flip on negative determinant), bbox, tessellation, shell and degraded booleans all handle it. The evaluator's *"no path from a triangle soup to a BRep solid"* comment was simply stale.

**`vcad info` on XLeRobot: 144 triangles → 106,480.** STEP export still correctly errors `NotBRep`; the render path is byte-identical (verified by checksum — it takes precedence via `find_imported_mesh`).

### Links with interleaved `<visual>`/`<collision>` failed to parse

quick-xml's serde adapter only folds *contiguous* repeated siblings into a `Vec`. XLeRobot's `base_link` is collision×4, visual×3, collision×4 — an ordinary convex decomposition — and died with `duplicate field \`collision\``. A normalize pass already existed at the `<robot>` level for exactly this reason; it now runs inside `<link>` too.

### Relative mesh paths were never anchored in the CLI or renderer

A document written by `import-urdf --relative-meshes` — the mode the importer *recommends* for anything committable — resolved only when invoked from precisely the right working directory, and otherwise produced an empty page plus a pile of \`cannot open\` lines. Fixed at the central loaders (\`load_doc\`, \`load_vcad_document\`, vcad-render's \`read_document\`).

### Actions and observations index joints differently

Actions have one entry per **actuated** joint; observations have `max(1, ndof)` entries over **all** joints, Fixed ones keeping a zero slot. So `joint_positions[action_index]` silently returns a *neighbouring joint's* angle. It reads as a plausible tracking error — every joint appears to settle one neighbour off — rather than as an index bug, and a uniform action vector (any hold-pose baseline, or a `vec![5.0; n]` probe) hides it completely. That is why the K1 work never hit it. Adds `RobotEnv::joint_observation_index`.

## On the asset itself

**The URDF declares `effort="0"` on all twelve arm joints, so the arms are inert as imported.** Measured: a **40 N·m** command moves the `Elbow` by exactly **0.0000°**. Upstream never notices because ManiSkill supplies actuator limits from its own controller config and never reads the URDF's.

The importer now warns on any such joint but does **not** invent a replacement — an effort limit is a declared physical claim. `xlerobot_reach` supplies the Feetech STS3215's real 12 V stall torque of **2.94 N·m**, corroborated by upstream's own `gripper_force_limit = 2.8` (same servo, the one place they wrote a physical figure). Deliberately not adopted: their `arm_force_limit = 250` and `stiffness = 2e4`, which are SAPIEN drive numbers — 250 N·m on a 148 g forearm is not a servo.

Result, and a control to show the torque budget is real:

| servo | worst steady-state error over 6 joints |
|---|---|
| 2.94 N·m (STS3215) | **0.000°** |
| 0.02 N·m (starved) | **90.3°** |

Two caveats are documented rather than silently corrected, in `third_party/xlerobot/README.md`: `base_link` is authored at **70.13 kg** against a real robot of ~10 kg (left as upstream wrote it, so vcad and ManiSkill agree on base acceleration), and eleven links carry no `<inertial>` — all of them pure sensor/TCP frames on fixed joints, which is ordinary URDF and not a defect.

Gains are left to vcad's default inertia-scaled PD law (`kp = I·ω²`), which is stable by construction — these links span a 45× inertia range from Elbow to Jaw, and any single hand-picked `kp` that tracks the Elbow is 10× past the stability limit at the Jaw.

## Testing

- Six new regression tests: interleaved visual/collision parsing (plus one asserting nothing is *dropped* by the reorder), zero-effort preservation, mesh-under-transform resolution, that the transform is actually applied, and both index-mapping tests.
- `cargo test --workspace` green apart from `mecheval-grader`'s `cube_in_void_falls_under_gravity`, which **fails identically with this branch stashed** — pre-existing, not from these changes.
- `cargo fmt --all --check` clean; clippy clean on every touched crate (the one remaining workspace error is in `stepperoni`, untouched here).
- `cargo check --target wasm32-unknown-unknown -p vcad-eval` clean — the wasm arm keeps the old no-geometry behaviour, since the browser flow rewrites these nodes to `ImportedMesh` before evaluation.

## Follow-up, not in this PR

Multi-collision links are parsed but still unused: the importer takes geometry from `<visual>` only, so `base_link`'s 8-piece decomposition is still represented by one mesh. Filed separately.


Only1MrAnderson:
WILD

ecto:
Thanks — all three addressed in 87bf56ba. There were no inline threads to reply on, so summarising here.

**Major · `MeshImport` load failure swallowed in `evaluate_mesh_leaf`** — documented rather than changed, because the `.ok()` defers the decision rather than discarding it. `eval_instance` is where it gets made and it *does* fail closed: a part with no resolvable geometry and no authored `<inertial>` is a hard error there (`cannot derive mass properties`), since the alternative is inventing the mass properties the dynamics run on. Only a part that already has authored inertials may fall back to a placeholder collider. An unopenable reference is also an expected state, not a fault — a browser-flow URDF import keeps the raw URDF filename with no filesystem behind it. The comment now says all of that and points a future caller at `crate::stl::load_stl` for the real I/O error.

**Minor · reorder drops comments/whitespace in a link body** — this was already commented at the `continue`, but the note was thin, so I extended it with the fact that actually makes it safe. The one consumer that reads a comment is `commented_out_floating_joint`, and I checked: `read_urdf_from_str_with_options` passes the **original** `xml` to `apply_floating_base`, not `normalized`. So comment-dropping cannot affect floating-base detection. That invariant is now written down next to the code that would break it.

**Minor · `apply_actuator_profile` matches by exact id** — this one I promoted rather than filed as a nit, because the failure mode is nastier than "patched returns 0". Zero patched joints means all twelve arms keep the URDF's `effort="0"`, the reach runs to completion, and it reports a worst steady-state error around 90° — which reads as a physics or tuning result, not as twelve dead actuators. Exactly the class of silent-wrong-answer this PR is about elsewhere. It now errors and names the joints it could not find:

```
Error: "expected all 12 arm joints, found 10 — missing: Elbow, Jaw_2. This document
is not the XLeRobot import this example drives; without the actuator profile its
arms carry the URDF's effort=\"0\" and cannot move at all."
```

Verified against a document with two joints deliberately renamed. `cargo fmt` clean, clippy clean on the touched crates, 75 tests still green, and the reach still reports 0.000° nominal / 90.3° starved.

PR #791, opened 2026-08-11
fix(booleans): cone booleans through the apex plane no longer drop the wall band

ecto:
## What

Fixes the boolean family where a planar face's plane passes through a cone's apex — e.g. a box whose corner-at-origin edge lies exactly on the cone axis (torture rand-002/030/086/130 at identity pose). The result used to lose the entire conical wall band and leave the planar splits unsewn: open rim circles, 34–41% volume deficit.

**Torture: 13 bad-geometry cases now pass** (rand-002/030/051/069/086/130/186/187/206/207/314/349, chain-10), **0 regressions** on the full 752-case corpus.

Note: the original handoff described this as a razor-thin-cylinder (cyl-flat) failure; dumping the actual case specs (new `vcad-torture` `dump_case` example) showed all affected cases are cone-vs-cube with the cone axis on the box corner edge.

## Why it broke (4 root causes, all needed)

1. **SSI**: `plane_cone` returned `Empty` when the plane contains the apex — every marching sample lands at v=0 and is rejected. The true intersection is 0/1/2 straight rulings; now solved analytically (`A·cos u + B·sin u = −C`) → `Point` / `Line` / `TwoLines`.
2. **No conical ruling splitter**: cone faces silently ignored `Line` curves. Added `split_conical_face_by_ruling`, mirroring the legacy cylinder line splitter (dense canonical rim chains via `canonical_arc_points`; handles full bands, sectors, and frozen dense loops). Ruling splits are clipped at record time to the t-interval where the curve lies on **both** faces, so sub-bands the cut never reaches keep conforming rims (planar faces keep the full chord — a planar split must reach the face boundary).
3. **Freeze skipped cones**: `freeze_circle_loops` only froze cylinder-carried circles, so cone rims stayed analytic and conformed only by resolution coincidence — the same gap #758 closed for cylinders. Cone rims now freeze onto the canonical grid; trim recognizes frozen full cone bands (v-range test), the cone circle splitter gained a dense-band v-split path, and `tessellate_conical_face` renders sectors/frozen bands verbatim via a two-chain zipper.
4. **Grid routing**: `circle_on_cylinder_wall` (canonical-vs-raw ring router for planar hole rings) didn't recognize cone rims — caught as a rand-318 regression during development and fixed by extending it to cones.

## Reviewer notes

- Based on `claude/vcad-difference-boolean-bug-0b6b4e` (PR #789), so this PR targets that branch; retarget to `main` after #789 merges.
- New regression test `crates/vcad-kernel-booleans/tests/cone_axis_on_box_edge.rs` asserts **volume against analytic values** (2%), not just watertightness, per the curved-face-classification precedent.
- `cargo test --workspace`, `cargo clippy --workspace --exclude vcad-desktop -- -D warnings`, and `cargo fmt --all --check` are clean.
- Changelog entry included (`fix`).

### DIFF
--- a/changelog/entries/2026-08-11-cone-boolean-rulings.json
+++ b/changelog/entries/2026-08-11-cone-boolean-rulings.json
@@ -1,0 +1,9 @@
+{
+  "id": "2026-08-11-cone-boolean-rulings",
+  "version": "0.9.4",
+  "date": "2026-08-11",
+  "category": "fix",
+  "title": "Cone booleans: apex-plane rulings and frozen rims",
+  "summary": "Booleans against cones no longer drop the lateral wall when a face plane passes through the cone axis; cone rims now freeze onto the canonical grid like cylinders.",
+  "features": ["booleans", "kernel"]
+}
--- a/crates/vcad-kernel-booleans/src/freeze.rs
+++ b/crates/vcad-kernel-booleans/src/freeze.rs
@@ -62,25 +62,33 @@ pub(crate) fn freeze_circle_loops(brep: &mut BRepSolid, segments: u32) {
         }
         let twin = he.twin;
 
-        // Find the cylinder that carries this circle (own face or twin's).
-        let cyl_of = |hid: HalfEdgeId| -> Option<CylinderSurface> {
+        // Find the cylinder or cone that carries this circle (own face or
+        // twin's). Both carry rings perpendicular to their axis; all the
+        // machinery below needs is the axis direction and a point on it.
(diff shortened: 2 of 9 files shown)


ecto:
Addressed the Choji findings in 80c2a469: the `TwoLines` arm of `split_conical_face` now applies the second ruling across every sub-face of the first split (with a new test asserting 3 sectors from a direct `TwoLines` split), and `plane_cone`'s apex detection uses a relative tolerance scaled by coordinate magnitude. Full torture corpus unchanged (670 pass, 0 regressions); clippy/fmt clean.

PR #792, opened 2026-08-11
fix(booleans): land post-#789 boolean fixes on main (cone apex rulings, T-junction stitch, healed fallback)

ecto:
## What

PR #789 merged to main at 15:37 UTC, but several fixes kept landing on its branch afterwards — including PR #790 (T-junction stitching in `tessellate_brep`) and PR #791 (cone apex-plane rulings + frozen cone rims), which merged into the *branch* at 17:23, after the branch→main merge. Main never received them.

This PR brings the branch's remaining commits to main:

- **#791 — cone booleans through the apex plane no longer drop the wall band**: plane-through-apex ∩ cone now yields its rulings (was `Empty`); new conical ruling splitter; cone rims join the canonical-grid freeze machinery; canonical routing for cone-rim hole rings; review follow-up (both `TwoLines` rulings applied, relative apex tolerance). 13 torture bad-geometry cases fixed, 0 regressions.
- **#790 — stitch T-junctions closed in `tessellate_brep` output**.
- Healed-fallback trigger refinements and a `perf` change (tessellate each operand at most once per boolean), plus a torture baseline refresh from the CI scorecard.

## Reviewer notes

- Merges cleanly into main (verified with `git merge-tree`); no `packages/kernel-wasm` artifacts in the diff, so the wasm-refresh workflow on main remains the single writer.
- Everything here was already reviewed in #789/#790/#791; this PR only routes the commits to main.


ecto:
Addressed the Choji findings in c29be02d:

- **T-junction healing quadratic worst case**: candidates were already bounded to open-edge endpoints, but each edge tested all of them — added a sorted-axis prefilter (candidates sorted along x, each edge binary-searches its tolerance-padded slab), removing the O(T·C) worst case on badly cracked meshes.
- **mesh_signed_volume duplication**: now `pub` and re-exported from the crate root; both test files call it instead of carrying copies.
- **Triple-tessellation on the flagged path**: not changed — the code already tessellates each operand exactly once and shares them via `operands` (gated so the common accepted-result path pays nothing). `result.to_mesh` and `alt.to_mesh` tessellate two *different* solids (the B-rep result and the fallback result), both needed for the validity/agreement oracle, so there is no repeated tessellation of the same solid to eliminate.

Full torture corpus: 671 pass / 0 regressions; clippy/fmt clean.

Issue #822, opened 2026-08-19
kernel: union of lapped/stacked coincident-face plates — one mirrored link >400 s, its mirror 10 s


ecto:
Found building `hardware/rose-pro`. Unioning a link's sheet plates into one solid — where lapped and stacked plates touch face-to-face and fold corners overlap by exactly one thickness — sent `R_shank` past **400 s** while the mirror-image `L_shank` (same plates, y negated) took **10 s**. Same topology, same dimensions, only the sign of y differs, so this is a float-coincidence degeneracy in the coincident-face union path, not a shape cost.

Workaround in the generator (`rose/pro/rose_pro_loon.py`): emit one root per plate and never union plates. That is also what a STEP export / BOM wants, so it stuck, but the asymmetry is a kernel bug worth a torture case: two plates sharing a full face (lap), two plates sharing an edge strip (fold corner overlap of one thickness), at both y signs.

The plates are built with `hardware/rose-pro/plates.loon`; the old unioned-link form can be reconstructed from the per-link groups in `rose-pro.loon`.

Issue #821, opened 2026-08-19
kernel: plate boolean cost is not monotonic in corner radius (r=16, r=25 are 5-10x slower than r=20)


ecto:
Found building `hardware/rose-pro` (3 mm sheet plates = two overlapping boxes + four corner cylinders, a unioned bolt-circle of bores differenced out).

Whole-link evaluation time for the hipA link (3 plates, 26 bores) as a function of the plate corner radius, everything else identical:

| r (mm) | time |
|---|---|
| 8 | 6.6 s |
| 12 | 5.4 s |
| 16 | **23.8 s** |
| 20 | 5.2 s |
| 25 | **68.3 s** |

At r=25 the whole 108-root robot took ~20 min to render instead of ~2.5. This is not a cost curve — it's a degeneracy at particular radii. Hypothesis: at those radii the 32-segment corner-cylinder tessellation vertices land (near-)coincident with bore rim vertices or the box edges, and the boolean's split/weld stage goes pathological.

Repro: `hardware/rose-pro/plates.loon` `plate-x/y/z` + `bore-*`; the rose generator (`rose/pro/rose_pro_loon.py`, `corner_radius`) currently pins r=20 to stay in the fast trough — that workaround should not be necessary.

Builder: see `plates.loon` in this repo for the exact construction.

PR #823, opened 2026-08-19
feat(loon+eval): module lib path + content-addressed root-mesh cache (rose-pro 149 s → 0.8 s warm)

ecto:
## Motivation

`hardware/rose-pro/rose-pro.loon` — 108 roots, one 40 KB generated file that pasted a 100-line helper preamble ahead of the roots — took minutes to render, and touching one plate re-evaluated everything. Two deliverables: a module convention so the helpers live in their own file without a textual include, and a cache so unchanged roots are never re-evaluated.

## 1. Modules

loon already had the right primitive: a module is a `.loon` file whose top-level `let`s are its exports (`pub let` narrows), imported as `[use m]` / `[use m [a b]]` / `[use m :as x]`, evaluated **once in its own environment** exporting *values* (the closures), not text. What was missing was a lib path and a written-down convention that every consumer shares.

- **Resolution order** (identical in `vcad info`/`export`, `vcad-render`, and MCP `load_document`/`create_cad_loon`): next to the importing file → each dir in `$VCAD_LOON_PATH` → host-supplied in-memory map. A file beside the importer shadows a lib-path module. Reasons in [`crates/vcad-loon/src/modules.rs`](crates/vcad-loon/src/modules.rs); MCP's `composeLoonModules` reads the same variable.
- **rose-pro**: helpers → [`hardware/rose-pro/plates.loon`](hardware/rose-pro/plates.loon) (hand-maintained module: `plate-x/y/z`, `bore-x/y/z`, `act-x/y/z`); `rose-pro.loon` is roots-only with `[use plates [...]]`. The rose generator (`~/Developer/rose/pro/rose_pro_loon.py`, separate repo) emits the new form; its output is byte-identical to the hand conversion here. `reference/*.loon` are URDF mesh imports with no helpers — unchanged.
- Docs: a *Modules* pattern in the loon guide.

## 2. Cache (`vcad_eval::cache`)

Content-addressed cache of evaluated **root meshes**, on disk at `$VCAD_CACHE_DIR` / `$XDG_CACHE_HOME/vcad` / `~/.cache/vcad`, layout `geom/<kernel id>/<sha256>.mesh`.

**Key** = SHA-256 of: the root's IR subgraph renumbered in canonical post-order (node ids and names don't matter — operands come from a new **exhaustive** `CsgOp::child_ids()` / `child_ids_mut()`, so a new variant with an operand is a compile error, not a stale hit) · tessellation segment count · sheet-metal fold flag · any `VCAD_NO_*` kernel knob · **kernel identity** = `vcad-eval/<ver>/<target arch>/<source hash>`, where the source hash is a `build.rs` content hash of every `vcad-kernel*`, `vcad-ir`, `vcad-eval`, sibling `tang` source file and `Cargo.lock`. Editing a boolean changes the id and the cache starts a new directory; two checkouts at the same source share one. Arch is in the key because the torture baseline differs x86_64 vs aarch64.

**Never cached**: roots whose subgraph reads outside the IR (`StepImport`, `MeshImport`, `PartInstance` — `CsgOp::reads_external_data()`, also exhaustive), failed/panicked roots, empty roots. **Only the mesh** is stored (`Solid` has no serialized form), so a hit yields `solid: None`; STEP export, `--raytrace`, `--photoreal`, `--section` and clash detection bypass the cache. Writes are atomic (tmp+rename); a corrupt record is evicted on read. `VCAD_CACHE=0` / `--no-cache` opt out. Every run prints one stderr line: `cache: 107 hit, 1 miss, 1 stored (<dir>)`.

`vcad-render` asks for its own segment count (64 / 128 hi-res) so a hit is **pixel-identical** to a miss — the first attempt cached the 32-segment evaluate mesh and the warm PNG differed in cylinder shading; now baseline/cold/warm PNGs are byte-identical.

## Timings (rose-pro, 108 roots, `vcad-render --view iso --size 1400`, M-series laptop, release)

| run | wall | cache |
|---|---|---|
| before this PR | 149 s | — |
| cold (populates) | 149 s | 2 hit / 106 miss / 106 stored, 63 MB |
| **warm** | **0.81 s** | 108 hit |
| **one-plate edit** (corner radius of `L_trunkS_side`) | **3.3 s** | 107 hit / 1 miss |
| `vcad info` warm | 0.22 s | (cold 557 s — 228 s of it is `info`'s clash-detection pass, pre-existing) |
| `vcad export .glb` warm | 151 s → **0.85 s** | `write_glb` re-evaluated the document uncached — now goes through the cached evaluator |
| `vcad export .stl` warm | 646 s → **0.28 s** | the `floating_floor` print check is O(floors×tris)×6; skipped above 200k triangles with a note |

The "2 hit" on a cold cache is real: `L_ankle_bendtab` / `R_ankle_bendtab` are each `let`-bound twice by the rose generator (two ankle plates share a tag), so the second shadows the first and both roots render the same plate — a rose bug the cache surfaced; reporting it there.

## Also

- Filed the two boolean pathologies with numbers: #821 (cost non-monotonic in corner radius), #822 (union of lapped coincident-face plates, >400 s vs 10 s mirror).
- `hardware/rose-pro/README.md` refreshed (stale `vcad render` notes; the module layout).
- Pre-existing: `cargo clippy -p vcad-cli --all-targets` fails on `raycast.rs:140` (`field_reassign_with_default`, new clippy) on `main` too — not touched here.
- No `packages/kernel-wasm/vcad_kernel_wasm*` artifacts touched.

## Tests

`cargo test -p vcad-loon -p vcad-ir -p vcad-eval -p vcad-render -p vcad-cli` green; new: fingerprint (id/name-invariance, operand order, settings, external-data refusal), mesh record round-trip + corruption, disk cache, `evaluate_document` hit/miss/store counts, lib-path provider + chain, end-to-end `$VCAD_LOON_PATH` shadowing; MCP vitest: lib-path fallback + a rose-pro smoke through the `load_document` path (108 roots).


ecto:
Follow-up on the branch (520764c9): the rose generator landed its H2 pass mid-way (elbow/foot/hand links; 49→66 plates), so `hardware/rose-pro/rose-pro.loon` is regenerated — **137 roots / 4397 nodes**, same nine helpers from `plates.loon`, evaluates clean. The bendtab duplicate-name bug the cache surfaced is fixed in the generator (repeated tags within a link get `_2`), so the cold run now reports `0 hit, 137 miss, 137 stored` as it should. The timing table above was measured on the 108-root file; the per-root behaviour (warm ≈ 1 s, one-plate edit ≈ 3 s) is unchanged.

rose side (local commits on `~/Developer/rose`, not pushed): `pro/rose_pro_loon.py` emits `[use plates [...]]` and drops the pasted HELPERS; `render.sh` exports `VCAD_LOON_PATH=$HOME/Developer/vcad/hardware/rose-pro` so `pro/cad/rose-pro.loon` resolves the module through the lib path — verified with `vcad info` from the rose checkout.

PR #864, opened 2026-09-06
raytrace: wgpu 30, and the renderer moves to kosm-render

ecto:
## Summary
- **wgpu 23 → 30**, so a wgpu-30 surface (kosm-view) can hand the tracer its device.
- **The path tracer moves out of vcad.** The integrator, GPU pipeline, device-resident scene, per-pixel history and denoiser, environments and HDRIs now live in \`kosm-render\`; \`vcad-kernel-raytrace\` becomes a BRep client implementing \`kosm_render::Geometry\` over BRep faces, with the WGSL geometry module contract (client defines \`trace_scene\`, \`hit_normal\`, \`hit_tangent\`, \`hit_material_index\`, \`hit_orientation\`; client bindings 1..=5, renderer 0 and 6..=14). vcad stays a pure CAD kernel.
- **Materials**: compile against kosm-render's principled \`Pbr\` (Disney/OpenPBR parameterization); the IR's transmissive extensions (transmission, IOR, dispersion, attenuation, thin-walled) map onto it.
- **Tracer bugs fixed along the way**: torus quartic sign pairing (Ferrari), GGX f32 cancellation at retro incidence, gradient environment never uploaded, camera-invisible emitters, self-intersection epsilon not scale-free, analytic solves formed at the eye not the surface, stale history folded in as fresh.
- One light per bounce drawn by power; render only the requested rectangles; pixel filter re-exported.

## Merge blocker (why draft)
\`Cargo.toml\` points \`kosm-render\` at a local path in the kosm repo (ecto/kosm#5). Before merge this needs to become a published \`kosm-render = "0.1"\` (or a git dep) once that PR lands. A plain checkout will not build until then.


ecto:
Dependency wiring so this builds from a clean checkout (and so kosm can pull vcad by git):

- `kosm-render` is a git dep pinned by `rev` to kosm commit `5c42c28` (ecto/kosm#6) rather than a `branch`, so the lockfile does not churn on kosm merges. Bump the rev deliberately. After ecto/kosm#6 merges, re-pin to a kosm `main` sha.
- `tang`, `tang-la`, `tang-expr` are declared by registry version; the existing `[patch.crates-io]` keeps every local/CI build on `../tang`. Lockfile unchanged.
- `vcad-loon` pins `loon-lang` by git rev (`c12ef56`, loon `main`), patched to `../loon` for local/CI builds. Lockfile unchanged.
- Raytrace tests/example gained the `splats: None` field kosm-render's `Scene` grew.

Verified: `cargo build -p vcad-kernel-raytrace --features gpu`, `cargo test -p vcad-kernel-raytrace --features gpu`, `cargo build -p vcad-kernel-gpu`, `cargo fmt --check`.

Note: no CI run has fired for this PR (or for any pull_request event since Sep 2) — worth checking Actions billing/limits.

Issue #894, opened 2026-09-17
kernel: chained mesh booleans re-split coplanar caps without bound (no coplanar re-merge after mesh_csg)


ecto:
`mesh::csg::mesh_csg` splits each operand's triangles by the carrier planes of the other's triangles (AABB-localised) and never merges anything back. When operands share large coplanar faces — every prismatic part: all stator operands span z 11.1..17.1 — each boolean in a chain re-splits the accumulated cap triangles again.

**Measured on the rana-60 stator:** the authored 50-step union fold through the mesh path had not finished after 9 minutes (RSS 1.5 GB and climbing; 16 GB before `propagate_boolean` was fixed in #890). Even with only one mesh union followed by 8 mesh differences, the part is ~103k triangles where the analytic equivalent is ~14.7k. Related: #879, #822.

**Proposal:** a pass at the end of `mesh_csg` that groups kept fragments by carrier plane + connectivity, merges each connected coplanar region into its boundary loops (holes allowed), drops collinear boundary vertices not used by any other-plane triangle (or t-junctions open), and re-triangulates. Must preserve volume, watertightness (keep a region's original triangles if it fails to triangulate cleanly), orientation and determinism. Trap: `mesh_fallback` runs `QuadricCtx::project_mesh`, which reads per-vertex constraints off incident triangle normals — restrict the merge to exactly planar regions that are not quadric-projected.

**Done when:** a 12-step chain (ring, then 12 posts unioned one at a time through `mesh_csg`) ends within 3× the triangle count of the single-shot union, at the closed-form volume.

Found while getting the rana-60-cnc stator to solve (#890). Helpers `ring()`, `post()`, `fillet_block(side)` live in `crates/vcad-kernel-booleans/tests/coplanar_cap_union.rs`; assert volume against a closed form in any new test — every failure in this family was a plausible mesh of the wrong solid.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

ecto:
Partial work parked as draft #904: a coplanar-region re-merge that takes the one-shot ring ∪ 12 posts from 10 160 to 2 922 triangles and makes a 12-step chain's count fall instead of rise, at the closed-form volume with 0 open edges. It does not yet let the stator's authored 50-step fold finish — the draft's description has the measurements, the two regressions a less-guarded version caused (`rana_60c_shell`, `sheet_fold::u_channel`), and why the current gates are calibration rather than principle. The stator itself no longer depends on this path since #901 keeps it analytic.

[...]

Issue #893, opened 2026-09-17
kernel: mesh boolean fallback is not reproducible across processes (0 vs 12 open edges, 0.4% volume wobble)


ecto:
The stator's last big union goes through `api::mesh_fallback(.., refine = true)`. Same binary, same input:

- most runs: 8118.6 mm³, 0 open edges
- ~1 run in 3: 8150.6 mm³, 12 open edges (12 raw boundary edges)

**It only flips when no `VCAD_*` debug env var is set**: 0 of 26 runs with `VCAD_BOOLEAN_WARN`, `VCAD_UNION_TRACE` or a log-path var present, 5 of 14 without. None of those vars affects control flow (they only print), so the size of the environment block is what matters → address/alignment-dependent behaviour, uninitialised state, pointer-ordered containers, or float summation whose order depends on slice alignment. In-process repeats of `mesh_csg` + `remove_interior_membranes` + `repair_watertightness` on fixed operand meshes are bit-identical (8/8), so plain `HashMap` order does not explain it; `QuadricCtx::project_mesh` and the operand tessellation were not part of that check. Final triangle counts of the part also vary run to run (99 985 … 105 493).

**Why it matters:** fidelity and volume of a cached, content-addressed result depend on the run that populated the cache. #890 made the union referee tolerant (accepts a fallback ≥ 20× cleaner than the analytic result) so the verdict no longer flips, but the volume still wobbles.

**Method that should work:** unconditional (not env-gated — env vars mask it) stage hashes of operand meshes, raw CSG, each projection, membranes, repair; run ~10× under `env -i` with a padding variable of varying length; find the first stage that differs.

Found while getting the rana-60-cnc stator to solve (#890). Helpers `ring()`, `post()`, `fillet_block(side)` live in `crates/vcad-kernel-booleans/tests/coplanar_cap_union.rs`; assert volume against a closed form in any new test — every failure in this family was a plausible mesh of the wrong solid.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

Issue #897, opened 2026-09-17
kernel-gpu: two wavefront tests fail on Apple silicon (macOS 27)


ecto:
`cargo test -p vcad-kernel-gpu wavefront::tests::gpu_compacted` fails deterministically on an M-series Mac running macOS 27:

- `wavefront::tests::gpu_compacted_matches_cpu_on_walled_grid` (wavefront.rs:916)
- `wavefront::tests::gpu_compacted_matches_cpu_multi_source_random_obstacles` (wavefront.rs:946)

The other 19 tests in the crate pass, and CI (Linux, no GPU) is green. The crate has no dependency on the boolean kernel; noticed while running the full workspace for #890.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

PR #901, opened 2026-09-17
Kernel: unions of flush prismatic solids stay analytic, in either operand order

### LINKED ISSUE #891: kernel: union with a many-shell operand drops the coplanar caps over every overlap (stator falls back to triangle soup)
**Repro** (256 segments, all operands z 11.1..17.1): fuse 12 filleted posts (`post ∪ fillet_block(+1) ∪ fillet_block(−1)`, rotated 0..330°) into ONE multi-shell solid, then union with `ring()`.

| | volume | open edges |
|---|---|---|
| closed form | ≈ 7447 | – |
| `ring ∪ posts` (analytic) | 7360.8 (−1.16%) | 1358 |
| `posts ∪ ring` (analytic) | 7433.6 | 2448 |
| one filleted post at any of the 12 angles | 4950.4 vs 4949.9 ✓ | 6–38 hairline |

Rasterising the result's caps against the CSG reference shows the top AND bottom caps missing over every post∩ring region (r ≈ 24..25). With k = 2 or 3 posts the union already falls back (`VolumeDisagreement`).

**Consequence:** `api::boolean_op_reported`'s union referee (added in #890) overrules the analytic result, so the stator solves to the right volume (7831 vs 7848 mm³) but as ~103k-triangle soup instead of ~15k analytic triangles. `tests/coplanar_cap_union.rs::twelve_filleted_posts_into_the_ring` only passes because of the referee.

**Done when:** that test passes with `VCAD_NO_UNION_REFEREE=1` and both operand orders report `reason == None`.

**Lead:** `split_planar_face` takes `crossings[0]`/`crossings[1]` of the OUTER loop for every recorded segment of a line. On a face the line crosses in several spans (a notched plate, an annulus cut by many chords) it cuts the first span repeatedly and never the others — confirmed as the cause of the sheet-metal bend-relief discrepancy (fix: pass the trimmed segment down and pick the span it overlaps). The annular cap here is cut by ~60 full-chord lines.

Found while getting the rana-60-cnc stator to solve (#890). Helpers `ring()`, `post()`, `fillet_block(side)` live in `crates/vcad-kernel-booleans/tests/coplanar_cap_union.rs`; assert volume against a closed form in any new test — every failure in this family was a plausible mesh of the wrong solid.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

### LINKED ISSUE #892: kernel: union is operand-order dependent — `ring ∪ post` exact, `post ∪ ring` double-covers the overlap caps
**Repro:** `ring()` (r 24..28.75) and `post()` (8×5×6 cube reaching 0.5 mm into the ring wall), same z range so both caps are coplanar.

| | volume | net-open edges | over-used |
|---|---|---|---|
| closed form | 4946.55 | – | – |
| `ring ∪ post` | 4946.55 ✓ | 0 | 0 |
| `post ∪ ring` | 4951.99 (+5.44) | 30 | 24 |

+5.44 mm³ is exactly 2 × the overlap cap area (2.717 mm²): the caps over post∩ring (x ∈ [23.87, 24.5], |y| ≤ 2.5, at both z planes) are present twice.

**Established** (`--features debug-boolean`, `VCAD_SPLIT_DEBUG=1`): with the ring as operand B, its annular cap faces are offered the lines y = −2.5, y = +2.5, x = 16.5 and x = 24.5; every split returns "1 sub-faces" silently except x = 24.5. The overlap patch is never cut out of the ring's cap, the whole annulus is kept, and the post's own cap patch (kept as the A-side coplanar face) doubles it. Likely the same root as the many-shell issue: `find_line_polygon_crossings` only looks at the outer loop, so a chord entering and leaving the outer circle passes through the bore's inner loop and `split_face_by_curve` cannot partition the face.

**Done when:** both orders give 4946.55 ± 0.1%, closed.

Found while getting the rana-60-cnc stator to solve (#890). Helpers `ring()`, `post()`, `fillet_block(side)` live in `crates/vcad-kernel-booleans/tests/coplanar_cap_union.rs`; assert volume against a closed form in any new test — every failure in this family was a plausible mesh of the wrong solid.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

ecto:
Stacked on #899 (the `main` fix) — that commit disappears from this diff once #899 merges. Follow-up to #890.

## Result
The rana-60 stator (50-stage pipe of unions/differences, reference **7848 mm³** grid-integrated from its CSG source):

| | volume | fidelity | triangles | time (debug) |
|---|---|---|---|---|
| after #890 | 7830.6 (−0.22%) | TriangleSoup | 104 725 | 59 s |
| **this PR** | **7853.0 (+0.06%)** | **Analytic** | **11 709** | **24 s** |

Measured twice, same volume and fidelity both runs. The union referee added in #890 no longer has to overrule anything on this part.

## Three defects
1. **Coincidence judged one-sidedly** (`classify.rs`). A small cap patch contained in a large coplanar cap read `OnSame`, but the large cap read `Outside` against the patch — so with the small solid as operand A both copies of the overlap survived. `ring ∪ post` = 4946.55 (exact); `post ∪ ring` = 4951.99, 30 open edges; the +5.44 is exactly one doubled cap pair. New `FaceClassification::OnSameInner`: the larger face wins. "One-sided" is decided by whether the larger face actually reaches outside the smaller (boundary probes pulled 5% inward), not by area alone — a sheet-metal bend sector's end face is 1.3% larger than the flange face it sits on, and an area-only test dropped it (−1.6 mm³ per bend end). Both orders are now byte-identical.
2. **Phantom full-width chords** (`pipeline.rs`). Every post plane handed the ring's cap a chord across the whole face though a post reaches 0.5 mm into a 4.75 mm wall; two posts 30° apart put a chord crossing inside the cap and classification went incoherent (12 plain posts: −9.3%, 429 open edges; 12 filleted: −1.2%, 1358). A planar `Line` split is now skipped when the cutter face ends flush on that plane **and** the other solid has a coplanar, same-facing, *contained* face there. Both extra conditions are load-bearing: without "same-facing" a boss-on-plate loses the chords that carve its footprint (`bore_breaking_out_through_faces`), without "contained" a partially overlapping pair (sheet-metal flange over its plate) loses its only separator.
3. **Wrong span cut** (`split.rs`). `split_planar_face` used `crossings[0..2]` for every recorded segment of a line; on a notched face the line crosses in three spans and the first was cut three times. It now picks the span the trimmed segment overlaps (faces with < 4 crossings take the old path). This was the real cause of the bend-relief discrepancy noted in #890: the *relieved* solid kept a membrane the un-relieved one correctly dropped. Delta is exactly 162.0000 again and the tolerance in `loon_sheet_metal.rs` is back to `< 1.0`.

## Tests
New: `a_contained_coplanar_patch_does_not_double_the_larger_face`, `twelve_filleted_posts_stay_analytic` (asserts `report.reason == None` in both operand orders) in `tests/coplanar_cap_union.rs`; `relief_notches_are_the_only_delta_across_a_split_edge` in `vcad-kernel`. All assert volume against closed forms.

`cargo test -p vcad-kernel-booleans -p vcad-kernel -p vcad-eval -p vcad-torture -p vcad-ffi --no-fail-fast`: 548 passed, 0 failed (includes `shell_ring_manifold`, `torr_boolean_catalogue`, `fidelity_baseline` — no re-bless needed). Workspace clippy (1.98.1) and fmt clean.

## Still open
~100 unpaired edges on the stator and ~180 on the 12-filleted-post union, at the tangent fillet rims — analytic and within tolerance, not watertight. #893 (fallback reproducibility) and #894 (coplanar re-merge) are unchanged.

Closes #891, closes #892.


ecto:
CI caught a real regression in the first push: the phantom-chord gate was operation-agnostic, but its premise (the larger coplanar face survives whole) only holds for a union, so a flush through-hole kept its caps (`vcad-kernel-sketch` donut extrude: 1120.79 vs 1053.88 mm³). The gate is now union-only, with a regression for difference and intersection. Verified before this push with the whole workspace the way CI runs it (`cargo nextest run --workspace --exclude vcad-desktop --profile ci`): 4382 passed; the only 2 failures are the local-GPU wavefront tests tracked in #897, which pass on CI. Stator unchanged at 7854.46 mm³ / Analytic / 11 708 triangles.

🤖 Posted by [Claude Code](https://claude.com/claude-code)

PR #903, opened 2026-09-17
Kernel: make boolean results reproducible across processes

### LINKED ISSUE #893: kernel: mesh boolean fallback is not reproducible across processes (0 vs 12 open edges, 0.4% volume wobble)

ecto:
Closes #893.

The same document evaluated to a different (equally valid) mesh from run to run — triangle counts wandered 99 985…105 998 on the rana-60 stator, and its mesh-fallback union returned 8118.6 mm³ closed or 8150.6 mm³ with 12 open edges. In-process repeats were always bit-identical, and the flip only ever showed when **no** debug env var was set, which pointed everywhere except the actual cause: `HashMap` iteration order (seeded per process) leaking into geometry in three places.

| where | what depended on hash order |
|---|---|
| `sew::merge_nearby_vertices`, `repair::weld_boundary_vertices` | merged vertices were removed in `merge_map.keys()` order; a slotmap recycles freed slots LIFO, so removal order decides which keys later vertices get, and with them every id-based tie-break downstream (one 378-face union: 4067 / 4075 / 4083 triangles) |
| `tessellate::snap_boundary_rails` | kept the FIRST nearest open edge (strict `<`) from a hash-ordered list; a slit's two rails routinely offer exactly equidistant candidates |
| `tessellate::heal_t_junctions_pass` | candidates built from `counts.iter()`; ties survived in that order into `dedup_by_key` |

All three now iterate sorted — three small hunks, no behaviour change beyond which of two equally valid tie-break winners is chosen (now always the same one).

**Verification** — across processes, since in-process repeats never showed it: 10/10 stator evaluations identical (volume, triangle count, defect count) under `env -i` with the environment block padded 37→370 bytes; an instrumented run gave byte-identical FNV hashes at all 7 fallback stages across 5 processes. Whole workspace the way CI runs it (`cargo nextest run --workspace --exclude vcad-desktop --profile ci`): 4378 passed; the only 2 failures are the local-GPU wavefront tests in #897, which pass on CI. Workspace clippy (1.98.1) and fmt clean.

Worth a look in CI: the torture-track baseline is platform-specific and these tie-breaks now resolve one fixed way, so a case could legitimately move.


ecto:
Merged current `main` (which now includes #901) into this branch, since both change boolean tie-breaks and the first CI run predated #901. Verified the combination locally the way CI runs it: 4382 of 4384 pass (the 2 are the local-GPU tests in #897). The stator, evaluated three times under `env -i` with the environment padded to three different sizes — the condition that used to flip it — is now identical every run: 7852.96 mm³, Analytic, 11 708 triangles, 102 unpaired edges (before this PR the triangle count wobbled 11 708 / 11 709).

🤖 Posted by [Claude Code](https://claude.com/claude-code)

